PR #9, opened 2022-06-03
docs(aliasing): fix description of alias syntax

CITguy:
Aliasing docs claim that it uses JSON pointer syntax, but that is incorrect. The spec mentions that the community group is currently _researching_ the use of JSON pointer syntax for aliases, but [hasn't made it official](https://design-tokens.github.io/community-group/format/#aliases-references) (see screenshot).

PixelSnap 2022-06-03 at 09 49 30@2x

This PR is an attempt in updating the wording to reflect the _current_, dot-delimited path syntax.

### DIFF
--- a/www/src/pages/docs/tokens/index.md
+++ b/www/src/pages/docs/tokens/index.md
@@ -405,7 +405,7 @@ _Should a type be added? [Suggest it be added!](https://github.com/design-tokens
 
 ## Aliasing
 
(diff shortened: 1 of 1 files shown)


drwpow (approved):
Thanks for the correction. You know, I remember looking at that reference and thinking the same—that the alias syntax didn’t bear much resemblance to JSON pointers. I think I was confused by the spec wording but didn’t give it much thought. Your change is definitely clearer, thank you.

PR #11, opened 2022-10-13
Fix typings publishing in @cobalt-ui/core

ekursakov:
File `packages/core/src/@types/token.d.ts` is ignored by build => these types are missing in `dist` folder.
So when using `@cobalt-ui/core` as a NPM library half of types are not resolved.


packages/plugin-css/src/index.ts
```diff
@@ -227,7 +227,7 @@ export function transformTransition(value: ParsedTransitionToken['$value']): str
 export function transformTypography(value: ParsedTypographyToken['$value']): Record<string, string | number | string[]> {
   const values: Record<string, string | number | string[]> = {};
   for (const [k, v] of Object.entries(value)) {
-    values[kebabinate(k)] = Array.isArray(v) ? formatFontNames(v) : (v as ParsedTypographyToken['$value']);
```
ekursakov, line 230: Not sure what type should be used here, so just bypassed type checks for now.

drwpow: I believe that was the correct type?

drwpow:
Awesome! I’m in favor of this change. I’d just recommend reverting the type change that I think was correct. And once tests pass I’m happy to merge & ship this

packages/plugin-css/src/index.ts
drwpow: Oh I see the error you’re getting with that type change. Yes, this is fine.

This is some weirdness caused by the translation between all token types with different schemas vs a single token type with a fixed value. Normally TypeScript can discriminate based on `$type`, but when going back the other direction things get messy.

This is fine for now; I’ll clean this up later.

drwpow:
🤦 I forgot about that failing Figma test, sorry! Also need to fix that as well 😅.

Thanks for all your work on this one!

packages/cli/test/figma/paint.test.js
```diff
@@ -71,6 +71,6 @@ describe('paint', () => {
         {color: {r: 0.875, g: 0.40833336114883423, b: 0.14583337306976318, a: 1}, position: 0.7677738070487976},
       ],
     }),
-  ).to.equal('linear-gradient(-58.32deg, #f2e03a 21.02%, #df6825 76.44%)');
+  ).to.equal('linear-gradient(-58.32deg, #f2e03a 11.87%, #df6825 76.78%)');
```
ekursakov, line 74: I "fixed" this test, but it was same broken in `main` branch.
Seems to me it's a bug in the `linearGradient` function, which should also take `gradientHandlePositions` into account when computing stops.

drwpow: Yes this was a bit of some unfinished work I need to wrap up. Thanks!

drwpow:
Published in v0.6.3! 🎉

PR #19, opened 2023-02-11
Fix typeof check

bheston:
Fixed a `typeof` check that was failing when running in a browser, where `process` is undefined.

*I didn't get a PR template or checklist, but let me know if you want additional info*


drwpow (approved):
Thank you! Good catch

PR #22, opened 2023-04-12
Support new spec types

mike-engel:
This adds support for `fontName`, `fontWeight`, and `number` which are now present in the [newest spec](https://design-tokens.github.io/community-group/format/#types) (as of this writing at least!). It also drops support for `font` since it's not a discrete type any longer.

This also updates the tests for emitted typescript files, which weren't updated with https://github.com/drwpow/cobalt-ui/commit/ae8a4d68add68e8e019c8f4f3be794f778a422fa and were causing failures.


.changeset/config.json
```diff
@@ -6,5 +6,5 @@
   "access": "public",
   "baseBranch": "main",
   "updateInternalDependencies": "patch",
-  "ignore": ["@example/**", "cobalt-docs"]
+  "ignore": ["@example/**", "docs"]
```
drwpow, line 9: This is actually correct as it was; this scans for the name in `package.json`, not the dir name

docs/src/pages/docs/tokens/index.astro
```diff
@@ -109,11 +111,52 @@ const tokenDef = {
...
+  }
+}`}
+  />
+
+  <h2 id="font-weight">
+    <Token type="fontWeight" />
+    Font
+  </h2>
+
+  <p>A font weight as defined in <a href={tokenDef.fontWeight} target="_blank"><b>8.4</b></a>.</p>
```
drwpow, line 129: Great docs updates! 🎉 Thank you

examples/adobe/tokens.json
```diff
@@ -244,7 +244,7 @@
   },
   "font": {
     "family": {
-      "$type": "font",
+      "$type": "fontFamily",
```
drwpow, line 247: Thank you for updating all the examples! 🙏

packages/core/test/token.test.js
```diff
@@ -163,25 +163,43 @@ describe('8. Type', () => {
...
+    });
+
+    test('number', () => {
+      const json = {typography: {body: {$type: 'fontWeight', $value: 100}}};
+      const tokens = getTokens(json);
+      expect(tokens.find((t) => t.id === 'typography.body').$value).toEqual(json.typography.body.$value);
+    });
+  });
+
+  describe('8.5: Duration', () => {
```
drwpow, line 202: Great tests!

drwpow (approved):
This is great, thank you!

.changeset/config.json
mike-engel: Ah ok, reverted!

mike-engel:
Thanks for the quick review @drwpow! I'm not sure why the tests are failing in CI—they pass locally for me, so I may be missing something or maybe ci is mis-configured?

mike-engel:
@drwpow Thanks! Happy to see those changes to the declarations file—TS didn't seem to like them in the repo but I didn't want to change them in this PR 😄 Rebased with #24 now

drwpow:
I’m going to publish a new version with these changes in. It’s time to confess that the plugin version bump to `1.x` while the core/CLI was on `0.x` was accidental; it’s just annoying to undo. Since I’ve already “broken the seal” and published beyond 0.x, I’ll just make this the 1.x bump for everything.

Plugins will get a minor bump for this even though it’s technically breaking semver, and I’ll just include “breaking” warnings in the changelogs. Then everything will be on 1.x and we’ll go from there.

This is probably fine because I’ve been using Cobalt in production for a while now and there’s nothing unstable about it, other than the W3C Design Tokens spec it’s built on.

mike-engel:
Awesome, thanks so much for creating cobalt @drwpow! I'm sure we'll have more PRs to submit as we implement tokens into our design system, but I'm glad there's at least one tool that follows the spec verbatim (rather than making non-spec edits)

PR #28, opened 2023-05-11
Make meta fields optional for JS

mike-engel:
This adds a new option to the js plugin to disable `meta` from being generated. For projects where tree-shaking is non-existent or doesn't work well, the meta object takes up a lot of space and probably isn't used much.

I'm not in love with the implementation at all really, so if you have thoughts on better ways to approach this, I'm happy to hear them! This approach was the best way to preserve order (i.e. not throwing meta at the bottom), but maybe that doesn't matter at all.

### DIFF
--- a/.changeset/dirty-dogs-switch.md
+++ b/.changeset/dirty-dogs-switch.md
@@ -1,0 +1,5 @@
+---
+'@cobalt-ui/plugin-js': minor
+---
+
+Allow meta fields to not be generated for javascript builds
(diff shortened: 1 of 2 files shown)


mike-engel:
@drwpow as written, this shouldn’t be a breaking change since it will only not include it if `meta` is false

drwpow:
Ah, sorry! I see now. Misread that line.

drwpow (approved):

PR #29, opened 2023-05-23
Update changesets script, release job

drwpow:
## Changes

CI-only change; updates changesets changelog generation and adds a release.yml npm job


docs/src/pages/docs/plugins/js.md
```diff
@@ -46,306 +46,6 @@ const red = token('color.red.10');
...
 ```
 
-You’ll also be able to see any `$description`s specified in your IDE in the form of JSDoc. If using TypeScript, this will be fully typed and will throw a type error if a bad token ID is passed.
-
-In addition, you’ll also find the following exports:
-
-| Name | Type | Description |
-| :--- | :--- | :---------- |
-
-# @cobalt-ui/plugin-js
```
drwpow, line 56: huh that’s so weird can’t believe a bad RegEx caused this /s

docs/scripts/update-readmes.js
```diff
@@ -26,7 +27,7 @@ for (const [input, output] of Object.entries(updates)) {
   }
 
   const dest = fs.readFileSync(new URL(output, import.meta.url), 'utf8');
-  const parts = dest.split('---');
```
drwpow, line 29: This clearly does not work for markdown tables 🙃

PR #61, opened 2023-07-28
Replace better-color-tools with culori

drwpow:
[See note](https://github.com/drwpow/better-color-tools/issues/45). Culori is a better color science option, but more importantly, is more performant.

_Technically_ this increases the footprint of Cobalt by a few kB, but it’s still pretty stinkin’ small. And this runs in Node, so footprint isn’t nearly as much of a concern as production frontend code.

### DIFF
--- a/packages/plugin-css/test/border/want.css
+++ b/packages/plugin-css/test/border/want.css
@@ -22,17 +22,17 @@
 
 @supports (color: color(display-p3 1 1 1)) {
   :root {
-    --ds-color-gray: color(display-p3 0.329412 0.278431 0.254902);
-    --ds-border: 1px solid color(display-p3 0.05098 0.011765 0);
+    --ds-color-gray: color(display-p3 0.32941176470588235 0.2784313725490196 0.2549019607843137);
+    --ds-border: 1px solid color(display-p3 0.050980392156862744 0.011764705882352941 0);
   }
 
   [data-color-theme="light"] {
-    --ds-color-gray: color(display-p3 0.329412 0.278431 0.254902);
-    --ds-border: 1px solid color(display-p3 0.05098 0.011765 0);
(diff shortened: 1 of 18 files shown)


packages/plugin-css/test/border/want.css
```diff
@@ -22,17 +22,17 @@
 
 @supports (color: color(display-p3 1 1 1)) {
   :root {
-    --ds-color-gray: color(display-p3 0.329412 0.278431 0.254902);
-    --ds-border: 1px solid color(display-p3 0.05098 0.011765 0);
+    --ds-color-gray: color(display-p3 0.32941176470588235 0.2784313725490196 0.2549019607843137);
```
drwpow, line 25: 🤔 I guess all the rounding I was doing was slowing my library down for no good reason

packages/plugin-css/test/color/want.css
```diff
@@ -47,35 +47,35 @@
 
 @supports (color: color(display-p3 1 1 1)) {
   :root {
-    --ds-color-blue: color(display-p3 0.129412 0.545098 1);
-    --ds-gradient: color(display-p3 0.129412 0.545098 1) 0%, color(display-p3 0.909804 0.352941 0.678431) 100%;
+    --ds-color-blue: color(display-p3 0.12941176470588237 0.5450980392156862 1);
+    --ds-gradient: color(display-p3 0.12941176470588237 0.5450980392156862 1) 0%, color(display-p3 0.9098039215686274 0.35294117647058826 0.6784313725490196) 100%;
```
drwpow, line 51: FWIW I’m glad I maintained my own color science library for a bit. If no other reason than to be sure the colors were accurate. With thousands of ways to do things in culori, it’s so easy to produce the wrong colors when upconverting to P3, and it would take a while to realize it.

PR #63, opened 2023-07-28
Add `co bundle` command

drwpow:
Adds `co bundle` command for combining multiple schemas into one.

Run independently of `co build`, which by default **will not output a bundled tokens file**


packages/cli/bin/cli.js
```diff
@@ -145,20 +165,22 @@ async function main() {
...
       break;
     }
-    default:
+    default: {
       showHelp();
+      break;
+    }
   }
 
-  console.info(`  Done  ${time(start)}`);
```
drwpow, line 161: Rolls time output into previous console line. Now there’s not a `Done` line (which was a bit extraneous)

PR #68, opened 2023-08-10
fix: broken links

agnelnieves:
Fixed broken links for the doc-site where it was taking back to the homepage.


agnelnieves:
No probs! Good stuff btw!

PR #69, opened 2023-08-15
chore: update font weight token code snippet

sareg0:
## Changes
I noticed the font weight section of the docs repeated the font family code snippet. I updated it, borrowing the code snippet shown in the Design Tokens spec: https://design-tokens.github.io/community-group/format/#font-weight

As well as this I updated the type  to be string. I assumed that 'number' is not a valid value for the $type

## How to Review

Check to see if the docs update I made looks good and on-par with the Design token spec. 

Here is a screenshot of the updated docs, from running locally
Screenshot 2023-08-15 at 18 42 29

### DIFF
--- a/docs/src/pages/docs/tokens/index.astro
+++ b/docs/src/pages/docs/tokens/index.astro
@@ -131,7 +131,7 @@ const tokenDef = {
     <tbody>
       <tr>
         <td><code>$type</code></td>
-        <td><code>number</code> or <code>string</code></td>
+        <td><code>string</code></td>
         <td><b>fontWeight</b></td>
       </tr>
       <tr>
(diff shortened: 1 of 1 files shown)


drwpow:
Ah thank you! Yes the example was a bad copy–paste. But no `fontWeight` actually accepts a number (`500`) or a string (`medium`). [See format](https://design-tokens.github.io/community-group/format/#font-weight)

sareg0:
> Ah thank you! Yes the example was a bad copy–paste. But no `fontWeight` actually accepts a number (`500`) or a string (`medium`). [See format](https://design-tokens.github.io/community-group/format/#font-weight)

Is that the value though that takes the number? I was under the understanding that the $type should only the the string "fontWeight", as showing in the Design Tokens spec docs 
e.g.
```
 "$type": "fontWeight"
```

drwpow:
Oh my mistake! You’re right—I read it wrong. Another bad copy–paste. Looks good to me

drwpow (approved):

PR #78, opened 2023-08-17
Update deps

drwpow:
## Changes

Updates deps. No code changes.

## How to Review

- CI should pass


packages/utils/src/token.test.ts
```diff
@@ -0,0 +1,26 @@
+import {describe, expect, test} from 'vitest';
+import {getAliasID, isAlias} from './token.js';
+
+describe('getAliasID', () => {
```
drwpow, line 4: 😅 can’t believe I made it all this time not having tests for these

Issue #73, opened 2023-08-16
Nested object support in the js plugin


mike-engel:
For some cases, having a proper nested object for js tokens would be preferrable over dot-notation strings:

```js
// fine, but not optimal sometimes
{
  'colors.gray': '#123456',
  'colors.black': '#000000',
  'colors.white': '#FFFFFF',
}

// would be better in some instances
{
  colors: {
    gray: '#123456',
    black: '#000000',
    white: '#FFFFFF',
  }
}
```

Would you be open to a config option to output a nested object rather than dot-notation string accessors?

drwpow:
Oh yes! That’s a great idea. While I still feel that a flat array is a better default, I do agree that there are times when you just want to export a flat JS object.

And the nice part is, TypeScript-wise, you don’t even need to type it! It’ll just figure itself out 😄 

But anyway, yes, would be open to adding this

PR #79, opened 2023-08-18
Allow JS plugin to nest token output

mike-engel:
## Changes

This implements a solution for #73, allowing the output from the JS plugin to be a nested object rather than a flat object with key IDs in dot-notation.

This converts JS, TS, mode, and meta data to all be nested (see the test `want` files). By default, this mode will be off to preserve backwards compatibility and can be turned on with the `deep` key in the plugin config. I'm not attached to that name, so I'm happy to change it.

Some internal type information had to change a little bit, but not much, and a new (safer? maybe, not sure) `serializeTS` method was added for simplicity. Additionally, a new util `object.set` was added for naive deep setting of a nested object key safely. It doesn't cover all the cases that something like `lodash.set` would, but that's ok with the current implementation of the library, and it's documented.

## How to Review

Tests are the most revealing for final output, and the actual implementation is a few new functions and some refactorings to use them.


packages/plugin-js/src/index.ts
```diff
@@ -103,33 +142,48 @@ export default function pluginJS(options?: Options): Plugin {
...
+            if (options?.deep) {
+              setToken(js.modes, `${token.id}.${modeName}`, await transform(token, modeName), true);
+            } else {
+              js.modes[token.id]![modeName] = await transform(token, modeName);
+            }
+            if (buildTS) {
+              if (options?.deep) {
+                setToken(ts.modes, `${token.id}.${modeName}`, `${tokenTypes[token.$type]}['$value']`, true);
+              } else {
+                (ts.modes[token.id] as TSResult['modes'])[modeName] = `${tokenTypes[token.$type]}['$value']`;
```
drwpow, line 183: 😅 I’m so sorry you had to wade through the manual `.d.ts` file generation. I don’t think a lot of this code was written as well as I’d like to be easy to refactor / work on.

But this all looks fantastic 🙂

drwpow (approved):
👏 This looks great, thank you!

Also I’m not opinionated on the name either. `deep` makes sense to me!

I don’t have any feedback; this looks great 🙂

packages/plugin-js/src/index.ts
mike-engel: Honestly it wasn’t that bad, it was just very specific! You did most of the work with the serializer function 😄

drwpow: For a lot of codegen stuff, taking a more AST approach is “cleaner” in some ways. But you pay for it in other ways in terms of upfront timesink / wading through the land of obscure bugs, not to mention when you set it down for a while, the confusion of “wait how does this thing work again?”

String mashing is sloppy but it’s real easy to compare “output good” vs “output bad” and know exactly where to make the adjustment

PR #81, opened 2023-08-25
plugin-css: Fix fontFamily alias bug

drwpow:
## Changes

In `@cobalt-ui/plugin-css`, when you aliased a `fontFamily` token it would break. It was expecting an array of font names but in the previous step it may have been transformed to a CSS var.

## How to Review

- Tests added

### DIFF
--- a/packages/core/test/w3c.test.ts
+++ b/packages/core/test/w3c.test.ts
@@ -192,9 +192,22 @@ describe('8. Type', () => {
     });
 
     test('array', () => {
-      const json = {typography: {body: {$type: 'fontFamily', $value: ['-system-ui', 'Helvetica']}}};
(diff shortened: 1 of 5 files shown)


packages/core/test/w3c.test.ts
```diff
@@ -192,10 +192,23 @@ describe('8. Type', () => {
     });
 
     test('array', () => {
-      const json = {typography: {body: {$type: 'fontFamily', $value: ['-system-ui', 'Helvetica']}}};
+      const json = {typography: {body: {$type: 'fontFamily', $value: ['system-ui', 'Helvetica']}}};
       const tokens = getTokens(json);
       expect(tokens.find((t) => t.id === 'typography.body')?.$value).toEqual(json.typography.body.$value);
     });
+
+    test('alias', () => {
```
drwpow, line 200: This test passed; just confirmation it wasn’t in the core parser

PR #92, opened 2023-08-31
Fix plugin-js object keys, sort tokens

### LINKED ISSUE #91: Modes can't begin with numbers
I created a "fontSize" `dimension` with a variety of modes, but I was surprised to find out that mode names can't be prefixed with numbers

Screenshot 2023-08-28 at 5 15 08 PM

Is this expected? Or is this a bug?

drwpow:
## Changes

Fixes #91. JS-valid keys weren’t being properly sniffed out, which led to invalid code generation.

Also sorts token keys so they will appear in a more predictable order.

## How to Review

- Tests should pass


packages/plugin-js/test/border/want.js
```diff
@@ -14,14 +14,14 @@ export const tokens = {
 
 export const meta = {
   border: {
-    '_original': {
-      '$type': 'border',
```
drwpow, line 18: This is now cleaner—these didn’t need quotes

packages/plugin-js/test/nested/want.d.ts
```diff
@@ -13,57 +13,57 @@ export declare const tokens: {
...
-      20: ParsedColorToken;
-      30: ParsedColorToken;
-      40: ParsedColorToken;
-      50: ParsedColorToken;
-      60: ParsedColorToken;
-      70: ParsedColorToken;
-      80: ParsedColorToken;
-      90: ParsedColorToken;
-      100: ParsedColorToken;
+      '10': ParsedColorToken;
```
drwpow, line 40: Now numbers get quoted. This shouldn’t be a breaking change; just errs on the side of safety (esp. with zero-padded numbers like `00`)

PR #104, opened 2023-09-05
Add Style Dictionary conversion

drwpow:
## Changes

Adds `co convert` command that converts a Style Dictionary JSON into the W3C format (poorly).

It’s very early, and buggy, but better than nothing.

## How to Review

- CI should pass (tests were added 🥳)


packages/utils/src/token.ts
```diff
@@ -12,3 +13,12 @@ export function isAlias(value: unknown): boolean {
   if (typeof value !== 'string') return false;
   return ALIAS_RE.test(value);
 }
+
+/** get last segment of a token ID */
+export function getLocalID(id: string): string {
```
drwpow, line 18: This is such a random little utility I didn’t realize I’d be needing as often as I do—getting the last segment of an ID path.

packages/cli/src/build.ts
```diff
@@ -3,7 +3,7 @@ import co from '@cobalt-ui/core';
 import fs from 'node:fs';
 import {URL} from 'node:url';
 
-export async function build(rawSchema: Group, config: ResolvedConfig): Promise<ParseResult> {
+export default async function build(rawSchema: Group, config: ResolvedConfig): Promise<ParseResult> {
```
drwpow, line 6: No reason for this change other than simplification. 🤔 IIRC I was on a very short-lived “no default exports” kick but I don’t remember why now. I never wrote any reason down for this.

PR #114, opened 2023-09-09
Support arrays of shadows, inset property

### LINKED ISSUE #88: Allow arrows of shadows
@drwpow i definitely agree with your recommendation you made here to [support multiple shadows](https://github.com/design-tokens/community-group/issues/100#issuecomment-1012516875), my specific use case is leveraging designs from Material UI, whose shadow tokens are an [array of 3 shadows](https://mui.com/material-ui/customization/default-theme/)

"0px 2px 1px -1px rgba(0,0,0,0.2),
0px 1px 1px 0px rgba(0,0,0,0.14),
0px 1px 3px 0px rgba(0,0,0,0.12)"

It sounds like you were considering expanding your tool to allow arrays, is that still the case?

drwpow:
## Changes

Fixes #88. Allows arrays of shadows in a backwards-compatible way in anticipation of growing community support. (older versions of Cobalt should still work with plugins. And declaring shadows as an object works, of course, because that’s still technically the spec.

Also adds the addition of the `inset` keyword which also seemed to win over community support.

## How to Review


packages/plugin-css/src/index.ts
```diff
@@ -286,110 +279,169 @@ export function transformLink(value: ParsedLinkToken['$value']): string {
...
+      }
+      const width = isAlias(originalVal.width) ? varRef(originalVal.width, {prefix: options?.prefix}) : transformDimension(value.width);
+      const color = isAlias(originalVal.color) ? varRef(originalVal.color, {prefix: options?.prefix}) : transformColor(value.color);
+      const style = isAlias(originalVal.style) ? varRef(originalVal.style, {prefix: options?.prefix}) : transformStrokeStyle(value.style);
+      return `${width} ${style} ${color}`;
     }
     case 'shadow': {
-      return transformShadow(value as typeof token.$value);
+      let {value, originalVal} = getMode(token, options?.mode);
+      if (typeof originalVal === 'string') {
```
drwpow, line 363: So a big, annoying change in `plugin-css` was:

**Before**: all tokens would be resolved to CSS variables if possible (i.e. any tokens that were aliases) before they were transformed

**After**: the alias action is moved to each individual token’s transformer

The reason for this change is there’s a lot of nuance and complexity between token values and aliases. Specifically with composite tokens—you can’t really transform the token where some parts may be aliased and some parts may not be, unless you know what token type it is first.

As a result of this PR, the code _feels_ a bit messier. But overall is much more stable than it was before. As evidenced by removing hacks like `CSS_VAR_RE.test(value)` where we didn’t know if something had been transformed or not yet (which we should know; and now we do)

drwpow: Also, as a result of this change, **there are no more transform* methods for composite types.** The reason is that in order to work properly, composite transformers need access to not only the token itself, but which mode is being accessed, and other meta-level information.

So now, the `defaultTransformer()` just keeps all context. Though we can clean up some logic for base tokens, unfortunately, all composite token transformation has to happen in one giant transformer. Otherwise it’s not really separating concerns; you’re just passing a giant context object around between functions which is probably worse

drwpow: Anyways, TL;DR, this also came up as a blocker for #113 as well, but this is the solution (I think). Anyway, we’ll know once I merge this and take another crack at that.

packages/plugin-css/test/shadow/want.css
```diff
@@ -0,0 +1,20 @@
...
+:root {
+  --ds-shadow-base: var(--ds-shadow-simple);
+  --ds-shadow-inset: inset 0 4px 8px 0 #00000026;
+  --ds-shadow-layered: 0 1px 1px 0 #0000001f, 0 2px 2px 0 #0000001f, 0 4px 4px 0 #0000001f, 0 8px 8px 0 #0000001f, 0 16px 16px 0 #0000001f;
+  --ds-shadow-simple: 0 4px 8px 0 #00000026;
+}
+
+@supports (color: color(display-p3 1 1 1)) {
+  :root {
+    --ds-shadow-inset: inset 0 4px 8px 0 color(display-p3 0 0 0 / 0.14901960784313725);
```
drwpow, line 16: 🎉

packages/plugin-css/test/shadow/want.css
```diff
@@ -0,0 +1,20 @@
+/**
+ * Design Tokens
+ * Autogenerated from tokens.json.
+ * DO NOT EDIT!
+ */
+
+:root {
+  --ds-shadow-base: var(--ds-shadow-simple);
+  --ds-shadow-inset: inset 0 4px 8px 0 #00000026;
+  --ds-shadow-layered: 0 1px 1px 0 #0000001f, 0 2px 2px 0 #0000001f, 0 4px 4px 0 #0000001f, 0 8px 8px 0 #0000001f, 0 16px 16px 0 #0000001f;
```
drwpow, line 10: 🎉 🎉 🎉

PR #117, opened 2023-09-12
Fix plugin-js shadow array

drwpow:
## Changes

`@cobalt-ui/plugin-js` would err on arrays of shadows (#114). This fixes that.

## How to Review

- Tests should pass


packages/plugin-js/src/index.ts
```diff
@@ -54,13 +54,19 @@ export function serializeJS(value: unknown, options?: {comments?: Record<string,
...
-  if (typeof value === 'function') throw new Error(`Cannot serialize function ${value}`);
-  if (typeof value === 'object')
+  if (typeof value === 'string') {
+    return `'${value.replace(SINGLE_QUOTE_RE, "\\'")}'`;
+  }
+  if (typeof value === 'number' || typeof value === 'boolean' || value === undefined || value === null) {
+    return String(value);
+  }
+  if (Array.isArray(value)) {
+    return `[${value.map((item) => serializeJS(item, {indentLv})).join(', ')}]`;
```
drwpow, line 64: Also minor fix to indent in JS output: it was inserting extra chars before

packages/plugin-js/src/index.ts
```diff
@@ -102,7 +114,7 @@ function defaultTransform(token: ParsedToken, mode?: string): (typeof token)['$v
   return {...(token.$value as typeof modeVal), ...modeVal};
 }
 
-function setToken(obj: Record<string, any>, id: string, value: any, nest = false) {
+function setToken<T extends Record<string, any>>(obj: T, id: keyof T, value: any, nest = false): T {
```
drwpow, line 117: This type change is irrelevant; just passes inference from the original object through better. But has no impact on the current code.

PR #127, opened 2023-10-06
Make the blockquote component visible on the dark scheme

aitorllj93:
## Changes

_What does this PR change? Link to any related issue(s)._

Before the change:

image

After the change:

image

## How to Review

_How can a reviewer review your changes? What should be kept in mind for this review?_

PR #130, opened 2023-10-16
Docs: semantic markup

### LINKED ISSUE #124: Docs blockquotes accessiblitly concerns
`<blockquotes>` are being used throughout the docs to provide call out notes. There are a number of them, but here are a couple for the sake of example.

- [the note about converting style-dictionary files on the getting started page](https://github.com/drwpow/cobalt-ui/blob/9fe22deff57c361585685da71bc513f4463db237/docs/src/pages/docs/getting-started/index.astro#L42C6-L42C16)
- [the note about file extensions on the config api reference page](https://github.com/drwpow/cobalt-ui/blob/9fe22deff57c361585685da71bc513f4463db237/docs/src/pages/docs/reference/config.md?plain=1#L34)

There are two accessibility issues with the current implementation.

**Contrast**

The white text on the light blue background is really not perceivable. I'm measuring a contrast ratio of 1.16:1, but text that size should be at least 4.5:1. This is a major issue for any sighted user. It looks like the background color is set in [styles/_md.scss](https://github.com/drwpow/cobalt-ui/blob/9fe22deff57c361585685da71bc513f4463db237/docs/src/styles/_md.scss#L57). I'm not a designer and don't really have a preference as to how this gets solved, but one easy way to do it would be using a darker blue for the background color. `blue-50` gets us up to 6.26:1 with the body text, but still leaves links below the threshold at 3.31:1.

**Semantics**

While not as big of an issue as the first, the second is the misuse of the `<blockquote>` element for styling purposes. This is unfortunately pretty common in docs sites, but `<blockquotes>` should only be used if the enclosed text is actually a quotation. Better options here would be to either use [the `<aside>` element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/aside) or [the `note` role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/note_role) (and probably move the related styling to a class). i realize markdown doesn't have a built in method for either of those approaches like the way a `>` lets you add a `<blockquote>`, but it should be possible to write the correct HTML directly inside the markdown file.

dev-nicolaos:
## Changes

Fixes several spots in the docs where the `<blockquote>` element was being incorrectly used to apply styling. Along with #127 , this change fixes #124 

There was also one spot on the getting started page where a closing tag for an `<ol>` got moved way down the page, breaking the page's flow. This PR also fixes that issue by moving the closing tag back up to the end of the list.

## How to Review

Build the docs and this branch and observe that the styling of the affected docs looks identical. Use dev tools to inspect the underlying markup and observe that the content changed now uses the correct semantics.

Fixing the issue with the closing `<ol>` caused prettier to change the indentation on much of the getting started page. It may be easier to review if you turn off showing white-space changes.

### DIFF
--- a/docs/src/pages/docs/getting-started/index.astro
+++ b/docs/src/pages/docs/getting-started/index.astro
@@ -33,19 +33,20 @@ const title = 'Getting Started';
       Running the CLI
       <li><i>(Optional)</i> Adding Cobalt to CI</li>
     </li>
+  </ol>
 
-    <h2>1. Creating the <code>tokens.json</code> file</h2>
+  <h2>1. Creating the <code>tokens.json</code> file</h2>
 
-    <p>
-      Converting your tokens to JSON is a manual process of grabbing values from your design files and formatting the JSON yourself. There’s not a great UI for this yet, but <strong>the Cobalt team is developing one currently</strong>.
-    </p>
+  <p>
+    Converting your tokens to JSON is a manual process of grabbing values from your design files and formatting the JSON yourself. There’s not a great UI for this yet, but <strong>the Cobalt team is developing one currently</strong>.
+  </p>
 
-    <blockquote><p>✨ Cobalt can now convert some Style Dictionary files into the W3C format (<a href="/docs/guides/style-dictionary">docs</a>)</p></blockquote>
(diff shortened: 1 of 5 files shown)


drwpow (approved):
Thank you!

PR #131, opened 2023-10-18
CSS Plugin: Handle Whitespace in Token Ids

### LINKED ISSUE #123: CSS Custom Props Have Spaces
It's possible for token or group object keys to contain a space. To the best of my knowledge, this is not disallowed by the spec (though specific cases like [the empty string](https://github.com/design-tokens/community-group/issues/195) and [leading/trailing white space](https://github.com/design-tokens/community-group/issues/194) are being considered).

Those space don't currently cause any issues (that I can tell) with Cobalt parsing the tokens, and the rendered output by the JS plugin is valid, using strings as keys. But they become a problem when trying to use the CSS plugin to produce custom properties. CSS custom properties cannot have spaces in them, but Cobalt renders the spaces as is anyway creating invalid CSS.

**Example** 

Tested with

- "@cobalt-ui/cli": "^1.6.0",
- "@cobalt-ui/plugin-css": "^1.5.0",
- "@cobalt-ui/plugin-js": "^1.4.1"

Given the following tokens input

```json
{
  "Full Color Palette": {
    "Magenta": {
      "Magenta-50": {
        "$type": "color",
        "$value": "#fff0f6",
        "$description": ""
      },
      "Magenta-100": {
        "$type": "color",
        "$value": "#ffd6e7",
        "$description": ""
      }
    }
  }
}
```

Running `npx co build` produces the following output:

<details>

<summary>A valid `index.js` file</summary>

```js
/**
 * Design Tokens
 * Autogenerated from tokens.json.
 * DO NOT EDIT!
 */

export const tokens = {
  'Full Color Palette.Magenta.Magenta-50': '#fff0f6',
  'Full Color Palette.Magenta.Magenta-100': '#ffd6e7',
};

export const meta = {
  'Full Color Palette.Magenta.Magenta-50': {
    _original: {
      $type: 'color',
      $value: '#fff0f6',
      $description: '',
    },
    _group: {
      id: 'Full Color Palette.Magenta',
      $extensions: {
        requiredModes: [],
      },
    },
    id: 'Full Color Palette.Magenta.Magenta-50',
    $type: 'color',
    $value: '#fff0f6',
    $description: '',
  },
  'Full Color Palette.Magenta.Magenta-100': {
    _original: {
      $type: 'color',
      $value: '#ffd6e7',
      $description: '',
    },
    _group: {
      id: 'Full Color Palette.Magenta',
      $extensions: {
        requiredModes: [],
      },
    },
    id: 'Full Color Palette.Magenta.Magenta-100',
    $type: 'color',
    $value: '#ffd6e7',
    $description: '',
  },
};

export const modes = {};

/** Get individual token */
export function token(tokenID, modeName) {
  if (modeName && modes[tokenID] && modeName in modes[tokenID]) return modes[tokenID][modeName];
  return tokens[tokenID];
}
```

</details>

<details>

<summary>An invalid tokens.css file</summary>

```css
/**
 * Design Tokens
 * Autogenerated from tokens.json.
 * DO NOT EDIT!
 */

:root {
  --Full Color Palette-Magenta-Magenta-50: #fff0f6;
  --Full Color Palette-Magenta-Magenta-100: #ffd6e7;
}

@supports (color: color(display-p3 1 1 1)) {
  :root {
    --Full Color Palette-Magenta-Magenta-50: color(display-p3 1 0.9411764705882353 0.9647058823529412);
    --Full Color Palette-Magenta-Magenta-100: color(display-p3 1 0.8392156862745098 0.9058823529411765);
  }
}
```

</details>

**Options**

1. This a known and expected limitation of the CSS plugin
    - If so, it would be a good idea to document it since this isn't an issue with other formats.
2. The plugin replaces any spaces in a key with a `-` the same way it does in between group and token names
3. The plugin replaces any spaces in a key with an `_` to distinguish between spaces within a name and the split between different token and group names
4. The developer is given the ability to handle this in the plugin's config. Maybe an expansion of the `transform` option to allow editing more than the value, or another option all together.

I think option 3 is my personal preference, but I don't have a strong opinion. Just hoping this can get resolved quickly.

Thanks for your work on this tool! Looking forward to seeing the design tokens space mature.

dev-nicolaos:
## Changes

This PR fixes #123 by removing leading and trailing white space from token and token group names and camelCasing those with white space in the middle of their names.

It also adds a new plugin option called `generateName` that allows the consumer to customize that behavior (and everything else about how a name is generated). This makes the `prefix` option redundant, so it has been deprecated. 

## How to Review

I broke the functions in the `util.ts` file into their own files in a `utils` folder for improved readability and test-ability, which made it a lot easier to work on the variable name conversion refactor in isolation. I also added tests for the bug fix and the new option. 

Observe that the unit and integration tests added for handling whitespace in token and group names pass.

The documentation added can be seen at the bottom of the options section of the plugin's page.


docs/src/pages/docs/integrations/css.md
```diff
@@ -82,6 +82,8 @@ export default {
       p3: true,
       /** normalize all colors */
       colorFormat: 'hex',
+      /** used to replace whitespace in the middle of group and token names */
+      spaceReplacement: '_',
```
drwpow, line 86: Thanks so much for this PR! But what are your thoughts on an alternative API that may be more flexible? It would let people specify different styles, e.g.

```suggestion
      variableName: (name) => camelcase(name)
```

I think we should still by default handle spaces in tokens. But in case people wanted to override that, they could pass in a function to enforce a specific style, e.g. `camelCase` or `snake_case` or `kebab-case` (all valid in CSS).

drwpow: Semi-related thought—I’m not super opinionated either on the default behavior, but I have noticed the W3C Design Token format prefers `camelCase` style when possible. Should we perhaps convert tokens with spaces to `camelCase` (and **only spaces**, to be clear)? e.g.

- `my-token` → `my-token` (same)
- `my token` → `myToken` (camelcased)

I’m just guessing at what would match the theoretical W3C styleguide the closest (which isn’t explicated, but just thinking out load)

docs/src/pages/docs/integrations/css.md
```diff
@@ -319,6 +321,29 @@ export default {
 };
 ```
 
+### Space Replacement
+
+Tokens and tokens groups can have whitespace in their names, but CSS variable names cannot contain whitespace. Leading and trailing whitespace will be removed during transformation and any whitespace in the middle of a group or token name will be replaced with the value of the `spaceReplacement` option.
```
drwpow, line 326: Great docs addition! Thanks for updating those. ❤️ (regardless of what the final API looks like, these are still very clear docs)

packages/plugin-css/test/build.test.ts
```diff
@@ -100,5 +100,22 @@ describe('@cobalt-ui/plugin-css', () => {
       });
       expect(fs.readFileSync(new URL('./actual.css', cwd), 'utf8')).toMatchFileSnapshot(fileURLToPath(new URL('./want.css', cwd)));
     });
+
+    test('spaceReplacement', async () => {
```
drwpow, line 104: 💯 great tests, too!

packages/plugin-css/src/utils/is-token-match.ts
```diff
@@ -0,0 +1,13 @@
+import wcmatch from '../wildcard-match.js';
+
+/** match token against globs */
+export function isTokenMatch(tokenID: string, globPatterns: string[]): boolean {
```
drwpow, line 4: Appreciate the extra attention to detail in cleaning up the utilities, and adding tests.

packages/plugin-css/src/utils/is-token-match.test.ts
```diff
@@ -1,5 +1,5 @@
 import {describe, expect, test} from 'vitest';
-import {isTokenMatch} from '../src/util.js';
```
drwpow, line 2: As a minor note, could you return the tests to the `/test/` folder? It makes it a little easier to exclude those from publishing to npm the way the package is set up. It also keeps all the tests together.

packages/plugin-css/src/utils/convert-token-name.test.ts
```diff
@@ -0,0 +1,53 @@
+import {describe, expect, test} from 'vitest';
+import {varName} from './convert-token-name.js';
+
+describe('varName', () => {
```
drwpow, line 4: Same here—really love this test, but if this were in the `/test/` folder along with the other tests, it’s easier to exclude it from npm (the way the package is currently set up)

packages/plugin-css/src/utils/is-token-match.test.ts
drwpow: aaaand I just realized I’m accidentally publishing plugin tests to npm! Will fix that in the next release. In that case folder doesn’t matter.

docs/src/pages/docs/integrations/css.md
dev-nicolaos: > what are your thoughts on an alternative API that may be more flexible? It would let people specify different styles

I could see that being useful to some people. Shouldn't be difficult to implement either. It does open up the possibility that the user can actually break middle space handling, but as long as we document it and handle it by default I think that's ok. Two questions on the proposed alternative.

1. You named the option `variableName` in your example.  This API will have to handle both variable and group names. Do you think that'll be understood by the user if we document it or should we come up with a different name?
2. Just double checking, with this approach we'd be running this option on every variable/group name right? Not just when we need to deal with a middle space?

> I’m not super opinionated either on the default behavior, but I have noticed the W3C Design Token format prefers camelCase style when possible.

> I’m just guessing at what would match the theoretical W3C styleguide the closest

Where have you noticed the `camelCase` style?  Looking through [the design tokens spec](https://tr.designtokens.org/format/) the few examples that match this issue all seem to use `kebab-case`.

packages/plugin-css/src/utils/convert-token-name.test.ts
dev-nicolaos: Resolving this thread to keep this conversation consolidated in [the first thread on this issue](https://github.com/drwpow/cobalt-ui/pull/131#discussion_r1367507831).

packages/plugin-css/src/utils/is-token-match.test.ts
dev-nicolaos: I think if you're trying to only ship what's necessary you could just specify `package.json` and the `dist` folder. Then anything that doesn't get built by `tsc` should remain out of the production bundle. The `exclude` line I added in `tsconfig.json` prevents these tests from ending up in `dist`.

That being said, if you still prefer they go into the `/test/` folder I'm can move them back no problem.

docs/src/pages/docs/integrations/css.md
drwpow: > Where have you noticed the `camelCase` style?

Mostly a historical change, from the standard moving from `kebab-case` to `camelCase` with things like `$type` changes ([PR](https://github.com/design-tokens/community-group/pull/111)). You’re right that most of the examples still haven’t changed from the original `kebab-case`, but the spec itself has decidedly chosen `camelCase`

packages/plugin-css/src/utils/is-token-match.test.ts
drwpow: No they’re fine where they are! I realized I’m publishing more to npm than I want to be, so I have cleanup that’s totally unrelated to your PR.

docs/src/pages/docs/integrations/css.md
drwpow: And to answer your 2 questions:

> 1. You named the option variableName in your example. This API will have to handle both variable and group names. Do you think that'll be understood by the user if we document it or should we come up with a different name?

I don’t feel too strongly on the name here. I was originally thinking it’d be easier to first generate a **valid CSS identifier** for each token output, but let the user override it, e.g.

```ts
pluginCSS({
  variableName((name) => {
    // var(--token-i-dont-like);
    if (name === 'token-i-dont-like') {
      return 'better-name';
    }
    return name;
  }),
})
```

Also, if we wanted to provide additional context, we could do something like:

```ts
// "token" is the original token
variableName((name, token) => {
```

But I’m open to other suggestions if you had a different API in mind! But overall I was thinking that with the addition of the `spaceReplacement` option, we’re providing a 2nd option to control naming **without letting the user actually control it** (the other being `prefix`). I’m anticipating other scenarios besides whitespace that a user may want to override a CSS variable name (e.g. perhaps they have some weird library interaction, or conflict, they need to handle) so turning this into a callback function would solve #123 while solving other future problems in one change.

> 2. Just double checking, with this approach we'd be running this option on every variable/group name right? Not just when we need to deal with a middle space?

That’s my thought, but again, open to pushback. I was thinking we’d pass back a **finished, valid CSS identifier** so we’d solve the #123 bug for users automatically. But let the user customize that if they wanted, for whatever reason.

drwpow: Love it! That’s exactly solving the problems I was thinking about, and then some. I agree with everything you proposed, and would like to implement it as-suggested. I only have a few minor thoughts which don’t change things too much:

- I would like to provide `token` as a 2nd parameter to `transformName`—that way users can choose to make different decisions, say, based on `$type` (most can ignore it, but I can think of a few scenarios where you may want context of the original token)
- Regarding “returning the full css variable name (`--` included)”, I’m fine with that but would probably also allow them to omit that if they’d like (i.e. add it if missing). Reason is I can’t imagine a scenario where you’d truly intend for `--` to be omitted? But happy to revisit if that’s not the case
- Regarding `prefix`: let’s keep for now so as we don’t have to release a breaking version of the plugin. But for v2.0 I’d drop `prefix` for sure and keep only `transformName`.

docs/src/pages/docs/integrations/css.md
```diff
@@ -76,12 +76,14 @@ export default {
       embedFiles: false,
       /** (optional) transform specific token values */
       transform: () => null,
-      /** (optional) add custom namespace to CSS vars */
+      /** (deprecated, use generateName instead) add custom namespace to CSS vars */
```
drwpow, line 79: Thanks for updating the JSDoc hints! I can make this change after merge, but I’ll put a `/** @deprecated` tag on it and that shows up in the IDE

docs/src/pages/docs/integrations/css.md
```diff
@@ -319,6 +321,52 @@ export default {
...
+        if (variableId === 'my.special.token') {
+          return "SUPER_IMPORTANT_VARIABLE";
+        }
+
+        return defaultNameGenerator(variableId);
+      },
+    }),
+  ],
+};
+```
```
drwpow, line 368: 💯 Awesome docs addition, thank you!

packages/plugin-css/src/utils/generate-token-name.test.ts
```diff
@@ -0,0 +1,74 @@
...
+    test('avoids duplicating leading dashes with prefix', () => {
+      expect(defaultNameGenerator('path.to.token', '--scope')).toBe('scope-path-to-token');
+    });
+
+    test('avoids duplicating trailing dashes with prefix', () => {
+      expect(defaultNameGenerator('path.to.token', '--scope-')).toBe('scope-path-to-token');
+      expect(defaultNameGenerator('path.to.token', 'scope-')).toBe('scope-path-to-token');
+    });
+  });
+});
```
drwpow, line 39: 👍

drwpow (approved):
This looks awesome! 💯 Sorry for the little bit of change there but I’m really happy with this solution, and it solves multiple problems in one PR. Thank you!

packages/plugin-css/src/utils/generate-token-name.ts
```diff
@@ -0,0 +1,49 @@
...
+export type CustomNameGenerator = (variableId: string, token?: ParsedToken) => string;
+
+// TODO: remove prefix arg in next major version
+export function makeNameGenerator(customNameGenerator?: CustomNameGenerator, prefix?: string) {
+  return (
+    /** dot separated path to token with possible type specific modifications (such as appended composite token property names)  */
+    variableId: string,
+    token?: ParsedToken,
+  ): string => {
+    const name = customNameGenerator ? customNameGenerator(variableId, token) : defaultNameGenerator(variableId, prefix);
```
drwpow, line 45: I like your implementation, but I will make one tiny change here in a patch release: we call a user’s `customNameGenerator()` on every token, and they’re allowed to return `undefined` or `null` if they’d like to defer naming back to us.

This is common plugin behavior ([example: Rollup](https://rollupjs.org/plugin-development/#a-simple-example)) and is a super minor detail that doesn’t affect the overall code. Just a tiny logical change that rather than checking a user passed the function, instead check the **return value** of the function and fall back to `defaultNameGenerator()` if `customNameGenerator()` executed but returned a falsy value.

packages/plugin-css/src/index.ts
```diff
@@ -507,20 +515,25 @@ function getMode<T extends {id: string; $value: any; $extensions?: any; _origina
...
 }
 
-/** convert token name to CSS variable */
-export function varName(id: string, options?: {prefix?: string; suffix?: string}): string {
-  return ['--', options?.prefix ? `${options.prefix.replace(DASH_PREFIX_RE, '').replace(DASH_SUFFIX_RE, '')}-` : '', id.replace(DOT_UNDER_GLOB_RE, '-'), options?.suffix ? `-${options.suffix.replace(DASH_PREFIX_RE, '')}` : ''].join('');
-}
-
 /** reference an existing CSS var */
-export function varRef(id: string, options?: {prefix?: string; suffix?: string; fallbacks?: string[]; mode?: string}): string {
+export function varRef(id: string, tokens: ParsedToken[], generateName: ReturnType<typeof makeNameGenerator>, suffix?: string): string {
```
drwpow, line 519: Also a note: will have to revert some of the changes to the `varRef` API as it’s used outside this plugin (may not be best practice, but it is what it is).

Mainly will revert the ordered params and revert back to a single object param for options (which is preferred in general, because then you can add/remove options easily, and handle backwards compatibility, with minimal effort)

packages/plugin-css/src/utils/generate-token-name.ts
drwpow: The breakage was really impossible to test for; it had to do with mismatching versions of plugin-css and plugin-sass. And since we didn’t indicate this as a breaking change, we shouldn’t require both be upgraded (I didn’t realize it either). Not really a way to test for this.

So note I’m not changing any code you wrote or how this function works; just going to release a `patch` for plugin-css and plugin-sass that changes some internals so both can work correctly (which, ironically, downloading the patch would fix the problem anyway, but it at least signals to consumers that a problem was identified and fixed).

Again, no one’s fault, not a big deal, and just sort of a weird, unpredictable, hard-to-test-for scenario. But wanted to leave a note.

packages/plugin-css/src/index.ts
```diff
@@ -345,88 +349,92 @@ export function transformStrokeStyle(value: ParsedStrokeStyleToken['$value']): s
   return String(value);
 }
 
-export function defaultTransformer(token: ParsedToken, {colorFormat = 'hex', mode, prefix}: {colorFormat: Options['colorFormat']; mode?: string; prefix?: string}): string | number | Record<string, string> {
+export function defaultTransformer(
+  token: ParsedToken,
+  tokens: ParsedToken[],
```
drwpow, line 354: Also here, too—didn’t notice this change to the `defaultTransformer()` function signature that was a breaking change for this plugin (because it taps into `transform()`). Again, there were no tests to catch this, so that’s a fault of the test suite! Just leaving notes.

PR #140, opened 2023-11-05
Docs v2

drwpow:
### Docs

CleanShot 2023-11-05 at 16 01 51@2x

- Moves the docs to [Vitepress](https://vitepress.dev/)
- Addition of docs search (thanks to Algolia)
- Lots of new content, better organization

While this looks a lot more “template-y” than the previous version (because it _is_), the first version of the docs were done as more of an experiment, and were a couple years old. I wanted to overhaul the content structure without investing time in a comprehensive redesign, and I rather like reading Vitepress docs anyway, so I focused on content updates over a visual refresh.

However, as my first time really using Vitepress I was blown away ❤️. There are so many amazing Markdown extensions built-in that make authoring much easier—things I would never build myself.

This doesn’t prevent Cobalt from having a custom docs site again down the line, but for now, I’d rather provide a good content & search experience rather than a bespoke design. And only move off the template when we outgrow it.

### Bugfixes

Also in writing the docs, makes some minor bugfixes:

- `@cobalt-ui/core` was missing the **object** style of `strokeStyle` tokens
- `@cobalt-ui/cli` was running `config()` in parallel, rather than synchronously before `build()` ran. While there weren’t any known problems, this just would have been frustrating to some plugin authors had they run into this.
- `@cobalt-ui/plugin-css`’ new `generateName()` API can return `undefined` to fall back to default naming

## How to Review

N/A


drwpow:
cc @dev-nicolaos you’ll be glad to know that Vitepress has [built-in callouts](https://vitepress.dev/guide/markdown#custom-containers) that don’t use `<blockquote>` 🙂. When reworking the content I wanted to preserve all the good a11y improvements you made.

While it’s not an `<aside>` anymore, at least it’s just wrapping `<p>` tags so it should follow normal document flow? But would love your thoughts (if you have any) on [how Vitepress handles callouts](https://vitepress.dev/guide/markdown#custom-containers).

dev-nicolaos:
That's an interesting feature. Couple thoughts:

- Paragraphs are certainly more appropriate than blockquotes. Whether the callout should additionally have a `complementary` (the `<aside>` element's role) or `note` role is going to depend on the content of each one. I think this is a two steps forward, one step back situation. 
- VitePress's callout implementation being a custom markdown syntax (`:::`  blocks) is an interesting choice. On the bright side, folks don't have to write HTML anymore. On the other hand, this makes the content less portable as I don't think this is commonly implemented by markdown parsers. The characters saved is somewhat minimal, and I wouldn't be surprised if the small time saved is lost by folks having to look up what that does and what values are available. Overall I dislike this implementation, but only a little.

Issue #145, opened 2023-11-06
CSS Plugin default transform should expose the same API and params with the custom one


radum:
Hello,

Thank you so much for Cobalt its been a life saving module so far :)

I have an issue with custom transforms when using the CSS Plugin. Since this [commit](https://github.com/drwpow/cobalt-ui/commit/6628a8cfe9adf20e9a6b437a82c7f6c1b28e1480) there have been some changes on how the internals work that could break or not bring the same flexibility for custom build transforms.

For example the current API for transform `transform(token, mode) {}` only has access to token and mode but if we want to reproduce the capabilities of the existing default transform we can't unless we reimplement some of the functionality ourselves.

For example looking at the typography one

```js
case 'typography': {
      const {value, originalVal} = getMode(token, mode);
      if (typeof originalVal === 'string') {
        return varRef(originalVal, tokens, generateName);
      }
      const output: Record<string, string> = {};
      for (const [k, v] of Object.entries(value)) {
        const formatter = k === 'fontFamily' ? transformFontFamily : (val: any): string => String(val);
        output[kebabinate(k)] = isAlias((originalVal as any)[k] as any) ? varRef((originalVal as any)[k], tokens, generateName) : formatter(v as any);
      }
      return output;
    }
    default: {
      throw new Error(`No transformer defined for $type: ${(token as any).$type} tokens`);
    }
  }
```

`varRef`

This needs new params that are not exposed to the `transform()` params. For example `tokens: ParsedToken[], generateName: ReturnType<typeof makeNameGenerator>` are needed and we can't make use of it.

We can get the list of parsed tokens via the core parser but that means we need to parse them again when they are already available.

And that is just one example.

I think the capabilities of the default transformer should be exposed in a reusable way to make sure we can reimplement the needed functionality.

There are more items to talk about this and I am wiling to help do that but just wanted to get your thoughts on it first.

drwpow:
Thanks for this feedback! And just to clarify, does the `config()` hook let you get access to the `generateName()` function as intended? Or is there something else needed?

For the `tokens: ParsedToken[]` bit could you speak a bit more about why you are transforming token A but need access to token B? Would want to understand why you’d need access to all the parsed tokens first before digging into a fix and/or workaround.

drwpow:
Ah sorry—yes there was an unintentional breaking change I’ll be releasing a patch for. I misinterpreted this as a need and not a bug. Will be releasing a patch to plugin-css and plugin-sass later today.

[...]

PR #150, opened 2023-11-06
Fix plugin-css, plugin-sass

drwpow:
## Changes

Fixes accidental regression in `transform()` APIs and plugin-sass ↔ plugin-css interop (mentioned in #145).

## How to Review

- Tests should pass.


packages/plugin-css/test/build.test.ts
```diff
@@ -96,6 +96,27 @@ describe('@cobalt-ui/plugin-css', () => {
       expect(fs.readFileSync(new URL('./actual.css', cwd), 'utf8')).toMatchFileSnapshot(fileURLToPath(new URL('./want.css', cwd)));
     });
 
+    test('transform', async () => {
```
drwpow, line 99: This API was missing a test; now it should catch regressions a little better.

drwpow:
🤔 Cloudflare failure seems to be a flake failure on their end (`Unknown internal error`). Seems unrelated to this PR.

packages/plugin-css/src/index.ts
```diff
@@ -518,25 +529,38 @@ function getMode<T extends {id: string; $value: any; $extensions?: any; _origina
...
+/**
+ * Reference an existing CSS var
+ *
+ */
+export function varRef(
+  id: string,
+  options?: {
+    prefix?: string;
+    suffix?: string;
+    // note: the following properties are optional to preserve backwards-compat without a breaking change
```
radum, line 541: Does this mean they will be remove with the `const token = options?.tokens?.find((t) => t.id === refID);` find and warn bellow once there is a breaking change?

drwpow: We’re following [semver](https://semver.org/) in this package, so there shouldn’t be any breaking changes until `2.0.0`.

This was just an unintentional internal change that accidentally led to a breaking change. It was triggered by mismatching versions of plugin-css (+/- `1.6.0`) and plugin-sass (+/- `1.3.2`). This change just allows for (hopefully) graceful fallback between the two plugins. The optional chaining method just allows an older version of the plugin to omit that data, and it will, still render a result.

drwpow: Hidden cross-dependencies like this are an antipattern in general, and something I should clean up better. Just at the moment, it felt like saving work / reducing bugs to have plugin-sass rely on plugin-css, but over time they’ve become too tightly-coupled and I think there needs to be a common core they share (irrelevant for consumers; this is just an implementation detail I’m musing on).

packages/plugin-css/src/index.ts
```diff
@@ -518,25 +529,38 @@ function getMode<T extends {id: string; $value: any; $extensions?: any; _origina
...
+    console.warn(`Tried to reference variable with id: ${refID}, no token found`); // eslint-disable-line no-console
+  }
 
   // suffix is only used internally (one place in plugin-sass), so handle it here rather than clutter the public API in defaultNameGenerator
-  const normalizedSuffix = suffix ? `-${suffix.replace(DASH_PREFIX_RE, '')}` : '';
+  const normalizedSuffix = options?.suffix ? `-${options?.suffix.replace(DASH_PREFIX_RE, '')}` : '';
   const variableId = refID + normalizedSuffix;
 
-  return `var(${generateName(variableId, token)})`;
+  return `var(${options?.generateName?.(variableId, token) ?? defaultNameGenerator(variableId, options?.prefix)})`;
```
dev-nicolaos, line 563: So does this mean `varRef` is a part of the plugin's public API? Because otherwise this is unnecessary. Both parts of this logic...

- Was a custom name generator passed? If not use the default one
- Was a prefix passed? If so, pass it to the default generator

...are already accounted for in the `generateName` argument (see line 76) by way of currying.

drwpow: >So does this mean varRef is a part of the plugin's public API?

Maybe not “public”, but used in the Sass plugin. And so mismatched versions would throw an error due to the changed signature. [It’s a TODO item for cleanup](https://github.com/drwpow/cobalt-ui/pull/150#discussion_r1385207498)

PR #173, opened 2023-12-19
Ensure alias conform example

autarc:
# Pull Request

## Changes

Currently the example shown at the ["Getting Started" guide](https://cobalt-ui.pages.dev/guides/getting-started#setup) is not working as the alias definition is not compatible. While the shortcut version of assigning an alias to any key directly is appealing, [it doesn't seem to be specified to that extent](https://design-tokens.github.io/community-group/format/#alias-reference) and is therefore not covered yet. To ease the introduction it is useful if first encountered example in the guides are working correctly as shown.

## How to Review

A test case was added to `build.test.js` which takes on the tokens definitions from the "Guides: Getting Started" section. It is run as part of the regular test suite. For the moment those are manually extracted but later on could be extracted to ease alignment.

## Note

While it would be convenient to reference a token as is with all its properties from `$value` to `$type` as it can reduce the amount of definitions - there might be pending questions in regards to granularity and composition. 

```json
"semantic": {
  "action": "{color.base.blue.5}",
  "textColor": "{color.base.gray.9}"
}
```

In the original example its clear that `color.semantic.action` refers to `color.base.blue.5` which is a color, but it could be that it is used as a `gradient` or other `$type`. Therefore in case aliasing gets extended to allow referencing all/multiple values from the selected JSON path of the tokens like an object rather than just a string - it might be an idea to have the closest `$type` available at the locale scope utilizes if it is available. Like the following:

```json
"semantic": {
  "$type": "gradient",
  "action": "{color.base.blue.5}",
  "textColor": "{color.base.gray.9}"
}
```

Otherwise if at the same level there is no `$type` specified it gets regularly applied from the referenced token. That said this behavior is currently not clarified in the specification so unless there is a larger set of repetitions it might be better for the time being to keep it restricted to single `$value` references only.

### DIFF
--- a/docs/guides/getting-started.md
+++ b/docs/guides/getting-started.md
@@ -121,8 +121,10 @@ color:
       '9':
         $value: '#002155'
   semantic:
-    action: '{color.base.blue.5}'
-    textColor: '{color.base.gray.9}'
+    action:
+      $value: '{color.base.blue.5}'
+    textColor:
+      $value: '{color.base.gray.9}'
 fontStack:
   $type: fontFamily
   sansSerif:
--- a/docs/guides/getting-started.md
+++ b/docs/guides/getting-started.md
@@ -52,8 +52,8 @@ Next, we’ll create a `tokens.json` file in the root of our project (or `tokens
       }
     },
     "semantic": {
-      "action": "{color.base.blue.5}",
-      "textColor": "{color.base.gray.9}"
+      "action": {"$value": "{color.base.blue.5}"},
+      "textColor": {"$value": "{color.base.gray.9}"}
     }
   },
   "fontStack": {
--- a/packages/cli/test/build.test.js
+++ b/packages/cli/test/build.test.js
@@ -50,4 +50,15 @@ describe('co build', () => {
       expect(Object.keys(builtTokens.meta).length).toBe(34);
     });
   });
+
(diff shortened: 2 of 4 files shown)


docs/guides/getting-started.md
```diff
@@ -52,8 +52,8 @@ Next, we’ll create a `tokens.json` file in the root of our project (or `tokens
       }
     },
     "semantic": {
-      "action": "{color.base.blue.5}",
-      "textColor": "{color.base.gray.9}"
+      "action": {"$value": "{color.base.blue.5}"},
+      "textColor": {"$value": "{color.base.gray.9}"}
```
drwpow, line 56: 🙏 Could you update the `YAML` example below as well?

drwpow (approved):
Ah thank you! Yes this was just a docs typo; that’s not intended behavior.

Thanks for the fix! If you could also fix the YAML example as well that’d be grand.

docs/guides/getting-started.md
autarc: Good point, updated it here https://github.com/drwpow/cobalt-ui/pull/173/commits/a5bebb504bfc0229921662fac4b7f2df67844bf5. The test suite foremost covers the JSON input so keeping it there as is.

PR #174, opened 2023-12-19
Remove reference of unsupported SASS plugin option

autarc:
# Pull Request

## Changes

Currently the instructions for [using CSS custom properties/variables with the SASS plugin](https://cobalt-ui.pages.dev/integrations/sass#css-variable-mode) are referring to using a `cssVars` option even though it is not necessary or even supported. Since its implementation (https://github.com/drwpow/cobalt-ui/commit/9edc9d9b4e8bbe2d1385dbbcc5476ef64e289ba1) the handling is covered by having `pluginCSS` set.


drwpow (approved):
Ah great catch. Thank you!

PR #175, opened 2023-12-19
Align optional Token transforms

autarc:
# Pull Request

## Changes

There is a mixed handling of the (optional) token transformers that can be applied throughout the plugins at the moment. While many share the same behavior there is an individual function signature specified along with internal referring to the respective default transformer. The later can differ based on the case, though for a consistent developer experience it is helpful to ensure an alignment which feels familiar if approaching multiple plugins. Therefore the base signature got added to the `core` package from which it gets referenced.

## Note

In addition to referencing the common function signature there were 2 other aspects considered:
* in some places the `options.transform?.(...)` return was declared to be asynchronous having it being `await`ed - since this was neither captured in the TypeScript function signature nor in the documentation it got omitted for consistency, if wished it might again be added as an alternative output to the synchronous processing
* the `transform` option for the Tailwind plugin wasn't documented yet, though it had a different signature referring to a record of `metadata` rather than the usual `mode` specifier, until it gets documented having it aligned with the others seems more helpful (even without the `mode` as its marked anyway as optional)

### DIFF
--- a/.changeset/modern-flowers-help.md
+++ b/.changeset/modern-flowers-help.md
@@ -1,0 +1,5 @@
+---
+'@cobalt-ui/core': patch
+---
+
+Define base token transform signature
--- a/.changeset/olive-trains-knock.md
+++ b/.changeset/olive-trains-knock.md
@@ -1,0 +1,8 @@
+---
+'@cobalt-ui/plugin-tailwind': patch
+'@cobalt-ui/plugin-sass': patch
+'@cobalt-ui/plugin-css': patch
+'@cobalt-ui/plugin-js': patch
+---
+
+Align plugin token transformers
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -74,6 +74,20 @@ export interface Config extends Partial<ParseOptions> {
   plugins: Plugin[];
 }
 
+/**
+ * Custom handling of token(s)
+ */
+export type TransformTokenFn = <T extends ParsedToken>(
+  /**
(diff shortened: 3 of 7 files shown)


drwpow:
Thank you for opening this! But I’m on the fence about these changes. While _in general_, yes, duplicating code is a small, in this instance these are separate plugins whose implementations of `transform()` are allowed to diverge at any time, and they don’t need to be bound by a central definition.

In other words, they are currently identical by _coincidence_, not _necessity_ (and the deviation Tailwind had you pointed out is allowed—and cleanup work I need to do either documenting it, or removing it).

Further, there are some implicit dependencies of `plugin-sass` on `plugin-css` that need to be removed (or abstracted more cleanly). In general, sharing types/methods/libraries between plugins is an antipattern I’d like to avoid. And since `transform()` isn’t part of the core plugin API, I think may also encourage other similar antipatterns of plugins sharing more than they should.

> in some places the `options.transform?.(...)` return was declared to be asynchronous having it being `await`ed -

This was a little bit of a hidden “user feature” even if a poorly-implemented one. The idea was that plugins should be synchronous, at least in their signature. But in the offchance someone _needed_ to run an async function, we’d wait for it. However, I didn’t want to ship an async type signature _encouraging_ async (slow) operations in plugins; merely allow them if someone needed. Congrats on being the first to discover it 😄 

---

For all the reasons above, I appreciate the cleanup, but I’m afraid I’ll have to vote against this PR landing (especially since this doesn’t fix any known bugs)

PR #196, opened 2024-02-16
Fix typography aliases

### LINKED ISSUE #195: Typography token aliases broken in mode-output of plugin-css
The [same bug I reported a few days ago](https://github.com/drwpow/cobalt-ui/pull/192) (thank you very much for the quick fix btw!) managed to hide away in another spot:

I'm afraid it is still happening within mode-output:

```
[data-viewport="xl"] {
  --fontalias-0: v;
  --fontalias-1: a;
  --fontalias-2: r;
  --fontalias-3: (;
  --fontalias-4: -;
  --fontalias-5: -;
  --fontalias-6: b;
  --fontalias-7: o;
  --fontalias-8: d;
  --fontalias-9: y;
  --fontalias-10: T;
  --fontalias-11: e;
  --fontalias-12: x;
  --fontalias-13: t;
  --fontalias-14: );
}
```

Here's the code to reproduce it:

tokens.json:
```
{
    "bodyText": {
      "$type": "typography",
      "$value": {
        "fontFamily": ["Helvetica", "-system-ui", "sans-serif"],
        "fontSize": "1.5rem",
        "fontStyle": "normal",
        "fontWeight": 400,
        "letterSpacing": 0,
        "lineHeight": 1.5,
        "textTransform": "none"
      }
    },
    "fontalias": {
        "$type": "typography",
        "$value": "{bodyText}",
        "$extensions": {
            "mode": {
                "xl": "{bodyText}"
            }
        }    
    }
}
```
tokens.config.mjs:
```
import pluginCSS from '@cobalt-ui/plugin-css';
/** @type {import('@cobalt-ui/core').Config} */

export default {
  tokens: ['./tokens.json'],
  plugins: [
    pluginCSS({
      modeSelectors: [
        {mode: 'xl', selectors: ['[data-viewport="xl"]']}
      ]
    })
  ]
};
```

Cobalt is a game-changer for our design system btw. Thank you!

drwpow:
## Changes

Fixes #195. And in the process fixes many other issues with typography aliasing that weren’t captured in the test suite. Overall a lot better.

## How to Review


packages/plugin-css/src/utils/token.ts
```diff
@@ -48,8 +48,8 @@ export function varRef(
   // suffix is only used internally (one place in plugin-sass), so handle it here rather than clutter the public API in defaultNameGenerator
   const normalizedSuffix = options?.suffix ? `-${options?.suffix.replace(DASH_PREFIX_RE, '')}` : '';
   const variableId = refID + normalizedSuffix;
-
-  return `var(${options?.generateName?.(variableId, token) ?? defaultNameGenerator(variableId, options?.prefix)})`;
+  const property = token && options?.property && typeof token.$value === 'object' && !!(token.$value as any)[options.property] ? options.property : undefined;
+  return `var(${options?.generateName?.(variableId, token) ?? defaultNameGenerator(variableId, options?.prefix)}${property ? `-${kebabinate(property)}` : ''})`;
```
drwpow, line 52: This is partially a hack, but only because of a hole in the DTCG spec. There’s currently no standard way to refer to **sub-properties** of a token, e.g. grabbing `fontFamily` from a `typography` token. Cobalt could extend the spec with this syntax, and it’s been suggested before, but I’m a little hesitant to make a decision on that because it feels imminent to happen in the spec this year.

Anyways, the “hack” is this doesn’t let people customize the final sub-property name even if they use the `generateName()` API, because there’s not an easy way to map any arbitrary string to a sub-property. Hopefully it’s not too big of a barrier for most folks.

packages/plugin-css/test/typography/want.css
```diff
@@ -17,7 +17,10 @@
   --ds-typography-family-body: "IBM Plex Sans", -system-ui, sans-serif;
   --ds-typography-family-footnote: var(--ds-typography-family-body);
   --ds-typography-family-heading: Helvetica;
-  --ds-typography-label: var(--ds-typography-body);
```
drwpow, line 20: This was actually broken and I didn’t notice. Now it’s not 😎

packages/utils/src/token.ts
```diff
@@ -3,11 +3,22 @@ import wcmatch from './lib/wildcard-match.js';
...
+  const rawID = match[1] ?? match[0];
+  const hashI = rawID.indexOf('#');
+  return hashI === -1 ? {id: rawID} : {id: rawID.substring(0, hashI), mode: rawID.substring(hashI + 1)};
+}
+
+/** @deprecated use getAliasValue instead */
+export const getAliasID = getAliasValue;
+
+/** Unwrap an alias value */
+export function getAliasValue(input: string): string {
```
drwpow, line 19: This is a little bit of a footgun here, and unofficially adopting the `#` character as a shorthand for modes has come back to bite me a little bit. There’s a little confusion over whether `#` is part of the ID or not.

Probably something to flag for cleanup in the next major release.

PR #205, opened 2024-03-03
Improve parsing of Tokens Studio files

### LINKED ISSUE #171: Possible bug with parsing Token Studio file
I have loaded up the example Token Studio file that is included in the plugin (also [on their repo here](https://raw.githubusercontent.com/tokens-studio/figma-plugin/b3b02db5d11c51d8e9ecad60ecd5df880cdbfa53/src/config/default.json)), however I'm getting some errors, which I'm guessing is likely due to the way the object is nested? The error is `scale: missing $type`

Is this a bug?

drwpow:
## Changes

Fixes #171, and also fixes several other Tokens Studio bugs in the process. While support isn’t perfect, it’s a lot better than it was, and will now be able to parse many more Tokens Studio exports that it would err out on before.

## How to Review

### DIFF
--- a/packages/core/src/parse/tokens-studio.ts
+++ b/packages/core/src/parse/tokens-studio.ts
@@ -3,8 +3,170 @@
  * This works by first converting the Tokens Studio format
  * into an equivalent DTCG result, then parsing that result
  */
-import {getAliasID, isAlias} from '@cobalt-ui/utils';
+import {parseAlias} from '@cobalt-ui/utils';
 import type {GradientStop, Group, Token} from '../token.js';
+
+// I’m not sure this is comprehensive at all but better than nothing
+const FONT_WEIGHTS: Record<string, number> = {
+  thin: 100,
+  hairline: 100,
+  'extra-light': 200,
+  extralight: 200,
+  'extra light': 200,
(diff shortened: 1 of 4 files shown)


packages/core/src/parse/tokens-studio.ts
```diff
@@ -24,28 +24,14 @@ export function convertTokensStudioFormat(rawTokens: Record<string, unknown>): {
...
+    for (const k in node) {
+      const v = (node as Record<string, unknown>)[k] as Record<string, unknown>;
       if (k.startsWith('$')) continue; // don’t scan meta properties like $themes or $metadata
 
       // token
       if (!!v && typeof v === 'object' && 'type' in v && 'value' in v) {
         const tokenID = [...path, k].join('.');
 
-        // resolve aliases
-        const tokenSet = path[0]!;
```
drwpow, line 35: TODO: figure out why this hack existed in the first place. Need to investigate whether Tokens Studio still does this—use pseudo-relative aliases or not (it’d be great if they didn’t any more)

drwpow: Figured it out lol

PR #208, opened 2024-03-05
Allow empty groups

### LINKED ISSUE #206: Allow empty groups?
I've noticed that Cobalt errors when it encounters an empty group in a DTCG file. In my current setup, which uses [Tokens Brücke](https://www.figma.com/community/plugin/1254538877056388290/tokensbrucke) to export DTCG from our Figma library's variables, I end up with an empty group. (FWIW, the reason we have an empty group is because our design system is still a work in progress, and we're scaffolding out groups before adding actual tokens into them.)

I've worked around this with a simple script that removes the offending group from the DTCG file, before it gets passed into Cobalt. However, I think it would be helpful if Cobalt was more permissive and just ignored empty groups rather than throwing an error. (or else this was downgraded to a warning)

It would simplify my setup a little. But I suspect it's a situation others might also encounter, so it could help others too.

FWIW, the [DTCG spec doesn't say anything about empty groups](https://tr.designtokens.org/format/#groups), so Cobalt's approach is a valid interpretation. However, I'll raise an issue in the DTCG repo later to suggest amending the spec to be more specific about whether (or not?) empty groups are allowed.

james-nash:
## Changes

Makes the parser log a warning instead of an error when it encounters an empty group. Closes #206.

## How to Review

Refer to the discussion in #206 for the rationale behind this change in behaviour.

I added a regression test that fails without this fix and passes with it. However, I'm not sure if the `expect(tokens.length).toBe(0)` line is really needed, as `getTokens()` already has an `expect()` assertion to check that there were no parsing errors. It felt a bit weird having a test without an `expect()` in it though, so I added that. Happy to remove it, if you think that's cleaner / clearer.

Since this is a bug fix, I've classed it as a patch change in the changeset.

### DIFF
--- a/.changeset/five-laws-stare.md
+++ b/.changeset/five-laws-stare.md
@@ -1,0 +1,5 @@
+---
+"@cobalt-ui/core": patch
+---
+
+Allows DTCG files to contain empty groups. Previously they would cause a parsing error, now only a warning will be logged.
--- a/packages/core/test/w3c.test.ts
+++ b/packages/core/test/w3c.test.ts
@@ -42,6 +42,15 @@ describe('5. Group', () => {
     const tokens = getTokens(json);
     expect(tokens.find((t) => t.id === 'color.blue')!.$type).toBe('color');
   });
+
+  test('allow empty', () => {
+    const json = {
+      color: {
+      },
+    };
+    const tokens = getTokens(json);
+    expect(tokens.length).toBe(0);
+  });
 });
 
 describe('7. Alias', () => {
--- a/packages/core/src/parse/index.ts
+++ b/packages/core/src/parse/index.ts
@@ -89,7 +89,7 @@ export function parse(rawTokens: unknown, options: ParseOptions): ParseResult {
         continue;
       }
       if (!Object.keys(v).length) {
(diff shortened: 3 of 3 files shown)


.changeset/five-laws-stare.md
```diff
@@ -0,0 +1,5 @@
+---
+"@cobalt-ui/core": patch
+---
+
+Allows DTCG files to contain empty groups. Previously they would cause a parsing error, now only a warning will be logged.
```
drwpow, line 5: 👍 Perfect! Agreed this should just be a patch.

packages/core/test/w3c.test.ts
```diff
@@ -42,6 +42,15 @@ describe('5. Group', () => {
     const tokens = getTokens(json);
     expect(tokens.find((t) => t.id === 'color.blue')!.$type).toBe('color');
   });
+
+  test('allow empty', () => {
```
drwpow, line 46: Nice test!

drwpow (approved):

james-nash:
Hurray! Thanks for the quick review & release. Much appreciated.

PR #211, opened 2024-03-06
Document meta option in JS plugin

### LINKED ISSUE #210: Option to exclude meta object from build output
Maybe I'm missing it, but from what I can tell the `meta` export in the output JS isn't used by anything else in the output. It does however, take up a huge amount of space. Our current build output's JS file is over 11,000 lines and the `meta` object is responsible for over 10,000 of them. This causes two issues:

- Unless you have ~~really sophisticated~~ (this behavior might be more common than I originally thought) tree-shaking happening at the sub-file level in your bundler, that makes token code dramatically more expensive to send to a browser.
- The updated dates in the metadata clutter the output when trying to review updates to the build output in Git (yes I know many workflows don't have include this, but ours does).

Since we aren't using the metadata, its a net negative for us. Having the option to exclude it from the build output would be really nice.

dev-nicolaos:
## Changes

Closes #210 

## How to Review

Observe that when the new test in `packages/plugin-js/test/build.test.js` is passed `meta: false`, the resulting output in `packages/plugin-js/test/meta/want.js` and `packages/plugin-js/test/meta/want.d.ts` does not include the `meta` object as this PR describes in the docs. This behavior can be contrasted with other tests in the same file that do not specify the `meta` option and have a `meta` object in their `want.*` files.


packages/plugin-js/src/index.ts
```diff
@@ -12,7 +12,7 @@ export interface Options {
   js?: boolean | string;
   /** output JSON? (default: false) */
   json?: boolean | string;
-  /** output meta? (default: false) */
+  /** include meta object in output? (default: true) */
```
dev-nicolaos, line 15: So this technically isn't true. There is no default for `meta` set anywhere, so the true default is `undefined`. But [the code explicitly checks for `false`](https://github.com/drwpow/cobalt-ui/blob/main/packages/plugin-js/src/index.ts#L161) so `undefined` has the same behavior as `true`. Given the option is semantically a boolean, I think its most helpful to say it defaults to `true` here and not bother developers with implementation details.

packages/plugin-js/test/build.test.js
```diff
@@ -45,4 +45,23 @@ describe('@cobalt-ui/plugin-js', () => {
...
+    await build(tokens, {
+      outDir: cwd,
+      plugins: [
+        pluginJS({
+          js: 'actual.js',
+          json: 'actual.json',
+          meta: false,
+        }),
+      ],
+      color: {},
```
dev-nicolaos, line 61: Its unclear why the tests fail without including this line in the call to `build` given that its not required for users to have it in their config files

packages/plugin-js/src/index.ts
drwpow: Good improvement!

packages/plugin-js/test/build.test.js
drwpow: This is the difference between **user config** and **resolved, normalized config.** In userland, we’re lax about config properties, and we fill in the gaps and provide defaults for anything not specified. But internally, we must fill in all defaults and normalize everything. This is a test of the internal build command which doesn’t want ambiguity; it wants all options filled in.

packages/plugin-js/test/build.test.js
```diff
@@ -45,4 +45,23 @@ describe('@cobalt-ui/plugin-js', () => {
...
+    });
+    expect(fs.readFileSync(new URL('actual.js', cwd), 'utf8'), `meta: JS`).toMatchFileSnapshot(fileURLToPath(new URL('want.js', cwd)));
+    expect(fs.readFileSync(new URL('actual.d.ts', cwd), 'utf8'), `meta: TS`).toMatchFileSnapshot(fileURLToPath(new URL('want.d.ts', cwd)));
+    expect(fs.readFileSync(new URL('actual.json', cwd), 'utf8'), `meta: JSON`).toMatchFileSnapshot(fileURLToPath(new URL('want.json', cwd)));
```
drwpow, line 65: Nice test!

drwpow (approved):
Awesome improvement, thank you! Great test!

PR #218, opened 2024-03-14
Add a11y testing plugin

### LINKED ISSUE #214: Contrast Tests (and others?)
## Proposal

Many developers use contrast tests either through annual audits, or E2E tests. And there are tools in Figma that alert of any contrast issues. But what if you could automatically run contrast checks just based on what’s in  `tokens.json`? What if you could, say, specify a foreground color, background color, and typography style, and have Cobalt test you’re meeting the minimum contrast ratios? Cobalt could even take modes into account as well!

The main idea is whenever you run `co build` it would fail with an error if tests didn’t pass (or could be a separate `co test` command; not sure). You could make as many token groups as you’d like, and assert whatever levels of compliance you need.

The simple contrast checker that most are familiar with is the simple WCAG 2 model which lets you naïvely compare any 2 colors and get a contrast ratio. But there are bigger, scarier contrast ratios like [APCA](https://www.myndex.com/APCA/) which not only give different scores based on which color is foreground and which is background, but also your font family, size, and weight used with those colors. Such a complex algorithm would be great to implement close to your tokens, especially when running in CI! 

#### 1. Config

The simplest implementation, e.g.:

```js
export default {
  tests: [
    {
      type: 'contrast',
      tokens: {
        foreground: '{color.semantic.text}',
        background: '{color.semantic.bg}',
        typography: '{typography.base}',
        modes: ['light', 'dark'],
      },
      expect: {
        apca: 'Lc 90',
        wcag2: ‘AAA',
      },
      severity: 'error', // 'error' | 'warn' | null,
    },
  ],
}
```

**Pros**: easy to test the idea, easy for users to quickly write tests

**Cons**: very limiting, as it won’t allow for “custom” user tests; either a test is built-in or not supported (so realistically there’d only be a couple test types)

#### 2. Lint plugins

Plugins get a new `lint` stage, which means they could accept any possible arguments for that stage (here, `checks`, but this could be named anything, and plugins could even reuse config you’ve already passed in).

```js
import contrast from '@cobalt-ui/plugin-contrast';

export default {
  plugins: [
    contrast({
      checks: [
        {
          tokens: {
            foreground: '{color.semantic.text}',
            background: '{color.semantic.bg}',
            typography: '{typography.base}',
            modes: ['light', 'dark'],
          },
          expect: {
            apca: 'Lc 90',
            wcag2: ‘AAA',
          },
          severity: 'error', // 'error' | 'warn' | null,
        },
      ],
    }),
  ],
}
```

Sure, the syntax seems similar to #1, but under-the-hood has lots of differences:

**Pros**: users can write custom tests, plugins can have additional “lint” awareness, and the tests could use more plugin-aware code (e.g. maybe in addition to your core `tokens.yaml` lint rules, you have additional linters that kick in when outputting CSS vars, such as warning on P3-gamut colors?)

**Cons**: if plugins try and be both “lint” and “build” plugins it can be a lot of work on the maintainer

#### 3. Testing

This ticket uses the word “testing” as a catchall, but in this option I mean _testing_-testing: like actually shipping a true testing package that executes inside a runner:

```js
// contrast-test.tokentest.js

import { assert, test } from 'node:test';
import { passesContrast } from '@cobalt-ui/testing';

const colorPairs = [/* (foreground, background, typography tokens) */];

for (const colorPair of colorPairs) {
  test("color combo passes accessibility", () => {
    assert.deepStrictEqual(passesContrast('./tokens.json', colorPair), { apca: 'Lc 90', wcag2: ‘AAA' });
  });
}
```

**Pros:** power, unlimited power!!!

**Cons:** this is woefully over-engineered, and would not be fun for me to maintain. Unit tests, by their nature, test runtime code. Tokens are statically-analyzable and there is no need to pretend they have runtime. This is not a serious option; I just wanted to write down an idea.

---

Of the three, I’m much leaning toward #2 as the option with the most potential, and could even be shipped in v1 without waiting for the [2.0 plugin API](https://github.com/drwpow/cobalt-ui/issues/201). Plus, kicking the tires now may inform how this system carries over into 2.x.

### Other tests?

Other than contrast, are there any other “tests” that would be helpful to run on your design tokens? Ideally ones that can be run programmatically and not require any heavy browser setup or anything.

We could have additional types like even token ID assertion, or that certain modes exist, as well.

drwpow:
## Changes

Fixes #214.

- Adds the ability for plugins to lint
- Adds `@cobalt-ui/lint-a11y` to test contrast

## How to Review

- Tests should pass

### DIFF
--- a/packages/lint-a11y/src/index.ts
+++ b/packages/lint-a11y/src/index.ts
@@ -1,0 +1,190 @@
+import { type Plugin, type ParsedToken, type LintNotice, type ParsedColorToken, type ParsedTypographyToken } from '@cobalt-ui/core';
+import { blend, modeRgb, modeHsl, modeHsv, modeP3, modeOkhsl, modeOklch, modeOklab, modeXyz50, modeXyz65, modeLrgb, useMode, wcagContrast } from 'culori/fn';
+import { APCAcontrast } from 'apca-w3';
+import { isWCAG2LargeText, round } from './lib.js';
+import { getMinimumLc } from './apca.js';
+
+// register colorspaces for Culori to parse (these are side-effect-y)
+useMode(modeHsl);
+useMode(modeHsv);
+useMode(modeLrgb);
+useMode(modeOkhsl);
+useMode(modeOklab);
+useMode(modeOklch);
+useMode(modeP3);
+useMode(modeRgb);
+useMode(modeXyz50);
+const toXyz = useMode(modeXyz65);
+
+export const RULES = {
+  contrast: 'a11y/contrast',
+};
+
+export const WCAG2_MIN_CONTRAST = {
+  AA: {
+    default: 4.5,
+    large: 3,
+  },
+  AAA: {
+    default: 7,
+    large: 4.5,
+  },
+};
+
+export interface RuleContrastOptions {
+  checks: RuleContrastCheck[];
+}
+
+export interface RuleContrastCheck {
+  tokens: {
+    foreground: string;
+    background: string;
+    typography?: string;
+    modes?: string[];
+  };
+  /** Enforce WCAG 2 contrast checking? (default: 'AA') */
+  wcag2?: 'AA' | 'AAA' | number | false;
+  /** Enforce APCA contrast checking? (default: false) @see https://www.myndex.com/APCA/ */
+  apca?: 'bronze' | 'bronze-body' | number | false;
+}
+
+function evaluateContrast(tokens: ParsedToken[], options: RuleContrastOptions): LintNotice[] {
+  const notices: LintNotice[] = [];
+  if (!Array.isArray(options.checks)) {
+    throw new Error(`"checks" must be an array`);
+  }
+  for (const {
+    wcag2 = 'AA',
+    apca = 'off',
+    tokens: { foreground: foregroundID, background: backgroundID, typography: typographyID, modes },
+  } of (options ?? {}).checks) {
+    // resolve tokens
+    const foreground = tokens.find((token) => token.id === foregroundID) as ParsedColorToken | undefined;
+    if (!foreground) {
+      throw new Error(`Couldn’t find foreground color "${foregroundID}"`);
+    }
+    const background = tokens.find((token) => token.id === backgroundID) as ParsedColorToken | undefined;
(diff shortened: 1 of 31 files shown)


packages/lint-a11y/src/index.ts
```diff
@@ -0,0 +1,237 @@
...
+};
+
+/**
+ * Table indexed by font size, then by weight ([100,200,300,400,500,600,700,800,900]).
+ * - First, start by finding the row where the the index is smaller than the given font.
+ * - Next, start with the font width, rounded DOWN to the nearest 100.
+ * A value of `null` means it’s innaccessible (out of readable range)
+ * @see https://www.myndex.com/APCA/
+ */
+export const APCA_FONT_LOOKUP_TABLE = [
```
drwpow, line 33: TODO: this still seems like a very “loose” guideline by APCA. There’s mention of “bronze,” “silver,” and “gold” compliance, but the actual guidelines based on font size + weight are pretty loose.

I was planning on the following workflow:

1. You (the user) have a color + typography set you want to test in APCA. You then specify `"bronze"`/`"silver"`/`"gold"` based on what you want.
2. If you pass it, great! Good job.
3. If you failed, the error message would say _by how much_ (e.g. `expected 50, got 49.8`).
  - If you wanted to fix it, you’d fix it
  - If for _reasons_ that’s good enough, you could “lock in” the value (e.g. round it down to `49`) and call it a day

However, even this is proving difficult to easily interpolate between the bronze and silver lookup tables (and I can’t find gold?), as the example tables and calculator only have partial data (and I’m not sure how they’re calculated).

Anyway, this is the only blocker: testing font size + weight. If it’s still a completely undecided area, I’ll punt and we’ll ship the initial version without font testing. But I plan on spending at least a little more time here before giving up.

drwpow: Found a little validation in the [WCAG 3 working draft](https://w3c.github.io/silver/guidelines/#conformance-levels) (emphasis added):

> **Gold** level identifies measures we want to include for those organizations that do achieve Silver so that some can stand out as exemplary, cutting edge, and role models. There are a number of ideas that **will be developed further once more of the conformance structure is solidified.**

PR #228, opened 2024-03-25
Improve gamut clamping

drwpow:
## Changes

- Adds 🆕 `color.gamut` option that lets you clamp to sRGB or P3 colorspace without rewriting your colors.
- Fixes `color.convertToHex` where it wasn’t properly gamut clamping (hex by definition requires sRGB space; there’s no such thing as P3 hex colors)

## How to Review

### DIFF
--- a/packages/core/src/parse/tokens-studio.ts
+++ b/packages/core/src/parse/tokens-studio.ts
@@ -3,7 +3,8 @@
  * This works by first converting the Tokens Studio format
  * into an equivalent DTCG result, then parsing that result
  */
-import { parseAlias } from '@cobalt-ui/utils';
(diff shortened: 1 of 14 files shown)


packages/core/src/parse/tokens-studio.ts
```diff
@@ -365,10 +369,65 @@ export function convertTokensStudioFormat(rawTokens: Record<string, unknown>): {
               tokenPath,
             );
           } else {
+            let value: string | undefined = v.value;
```
drwpow, line 372: This is pretty hacky, but fixes an existing bug where `rgba({colors.black}, 0.2)` would get left as-is.

Does a _little_ work to fix common scenarios, but is by no means exhaustive and there are still plenty of other broken scenarios I’m sure.

PR #230, opened 2024-03-26
Improve ESLint setup

drwpow:
## Changes

Improves ESLint setup. Parallelizes it (to also solve temporary OOM issues) and runs linting on more files.

## How to Review

- Tests (CI) updated; CI should pass

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
     "build:examples": "pnpm run -r --parallel --if-present build:tokens",
     "changeset": "changeset",
     "dev": "pnpm run -r --parallel --if-present --filter \"@cobalt-ui/*\" dev",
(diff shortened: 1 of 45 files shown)


package.json
```diff
@@ -13,7 +13,7 @@
     "build:examples": "pnpm run -r --parallel --if-present build:tokens",
     "changeset": "changeset",
     "dev": "pnpm run -r --parallel --if-present --filter \"@cobalt-ui/*\" dev",
-    "lint": "eslint \"packages/**/*.{ts,js,cjs,mjs}\"",
```
drwpow, line 16: There’s a current OOM bug where linting too many files from one process blows up. This splits the task into per-package lint rules (which will be helpful in the future if packages need additional linting)

PR #233, opened 2024-03-27
Improve APCA implementation

drwpow:
## Changes

🤔 There’s something still screwy with the APCA numbers I’m getting from `apca-w3` vs the [Contrast Checker Tool](https://www.myndex.com/APCA/). There do seem to be multiple different versions of the APCA formula floating around the interwebz, npm, and Myndex’ GitHub. Hopefully I can get a review soon of this implementation (requested)!

Right now I’m chalking up a slight variance of numbers to the `apca-w3` package possibly being newer? than some of the online tools. But perhaps there’s still a bug in my code somewhere. IDK, but either way I’m seeing _better_ numbers here with the fixes.

## How to Review

### DIFF
--- a/.changeset/dirty-points-sneeze.md
+++ b/.changeset/dirty-points-sneeze.md
@@ -1,0 +1,5 @@
+---
+"@cobalt-ui/lint-a11y": patch
+---
+
+Improve APCA implementation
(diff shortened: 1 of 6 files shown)


packages/core/src/parse/figma.ts
```diff
@@ -206,7 +209,7 @@ export function convertFigmaVariablesFormat(manifest: FigmaVariableManifest, opt
   }
 
   // 3. explode flat structure into nested structure
-  for (const [id, token] of Object.entries(dtcgFlat)) {
+  for (const id in dtcgFlat) {
```
drwpow, line 212: Here, and in other places, the performance crept up, so I removed a lot of array creation bloat with `Object.keys()` and `Object.entries()`.

It may not fix the slow test problem, but it should be a _lot_ more memory-friendly

PR #249, opened 2024-04-12
Fix bad lint test

drwpow:
## Changes

Chore; just fixes test which was giving false positives. Fortunately even though the _test_ was incorrect, the _code_ was working as intended so no patch/release needed 😅. But a future regression may have slipped through the cracks had this not been caught.

## How to Review

- N/A

### DIFF
--- a/packages/core/test/lint-core.test.ts
+++ b/packages/core/test/lint-core.test.ts
@@ -81,7 +81,7 @@ describe('rules', () => {
             },
           },
         },
-        want: { errors: ['Bad'] },
+        want: { success: true },
       },
     ],
     [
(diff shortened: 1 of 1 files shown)


packages/core/test/lint-core.test.ts
```diff
@@ -388,13 +388,11 @@ describe('rules', () => {
       rawSchema: given.tokens,
       rules: [{ id: given.rule, severity: 'error', options: given.options }],
     });
-    if (notices?.length) {
```
drwpow, line 391: This is actually not what we want—this relies too much on output and skips over the case where we _wanted_ an error but none was thrown.

Instead, prioritize the `want` cases first, and then assert the results match

PR #276, opened 2024-07-13
Add vertical slider, buncha other stuff

drwpow:
## Changes

Adds:

- `Slider` component to `@terrazzo/tiles` with `orientation: vertical` prop
- Fixes a pretty major bug in `@terrazzo/plugin-css` (hey, it’s still beta!)
- Reworks tokens
- Lots of other things I haven’t made up my mind on

## How to Review

- Lint is green
- Test is red
- Zendaya is Meechee


packages/react-color-picker/src/lib/webgl.ts
```diff
@@ -243,7 +243,7 @@ void main() {
   // 3 = projection toward point, hue dependent
   // 4 = adaptive Lightness, hue independent
   // 5 = adaptive Lightness, hue dependent
-  int clamp_mode = 3;
+  int clamp_mode = 2;
```
drwpow, line 246: Trying to find a good balance between weird Lightness & weird Chroma scales. 

Actual fix may be more involved, but just experimenting for now

PR #285, opened 2024-08-02
Support multiple tokens files

### LINKED ISSUE #283: Terrazzo not building json files into one token.json
Hi :)

We're migrating from Cobalt CLI to Terrazzo CLI, and trying to make the terrazzo config work correctly.

Currently we have an array of json files in the `tokens` property, and as the [docs](https://terrazzo.app/docs/cli/config/) say, this should be possible.
image

But when building, we can see that the tokens we get are only the ones from the first json file.

It seems like the build step from Cobalt CLI is not 1:1 with Terrazzo, or are we doing something wrong? :)

drwpow:
## Changes

Fixes #283. Adds multiple file support to Terrazzo CLI while retaining correct sourcemaps & filepointers for errors

## How to Review

- Tests added


packages/parser/test/parse.test.ts
```diff
@@ -283,15 +381,22 @@ describe('Tokens', () => {
...
+                  $value: [{ color: '{color.blue.500', position: '{perc.0}' }],
+                },
+              },
             },
-          },
+          ],
           want: {
             error: `Invalid alias: "{color.blue.500"
 
+/tokens.json:6:18
```
drwpow, line 398: This is the real magic: in error outputs, it preserves what file the error originated from, and also specifies line/column.

PR #293, opened 2024-08-19
fix(@terrazzo/token-tools): remove StringToken duplicated declaration in Token type

nclsndr:
## Changes

Removes the `StringToken` declared  twice within the `Token` union type in `token-tools/src/types.ts`.

## How to Review

I don't expect this change to have any side effect.


drwpow (approved):
Nice cleanup, thank you!

PR #295, opened 2024-08-19
draft(@terrazzo/token-tools): StrokeStyle dashArray allows dimension aliases

### LINKED ISSUE #294: Stroke style value should allow dimension aliases in dashArray
In DTCG specs we have the following [example](https://tr.designtokens.org/format/#example-stroke-style-obj-ref):

```json
{
  "notification-border-style": {
    "$type": "strokeStyle",
    "$value": {
      "dashArray": ["{dash-length-medium}", "0.25rem"],
      "lineCap": "butt"
    }
  },

  "dash-length-medium": {
    "$type": "dimension",
    "$value": "10px"
  }
}
```

Giving hint upon the fact the `dashArray` property can accept a dimension in [`StrokeStyleValueExpanded`](https://github.com/terrazzoapp/terrazzo/blob/main/packages/token-tools/src/types.ts#L195-L198)

```ts
export interface StrokeStyleValueExpanded {
  dashArray: DimensionValue[]; // Should become (DimensionValue | AliasValue)[]
  lineCap: 'round' | 'butt' | 'square';
}
```

nclsndr:
## Changes

Fixes #294

## How to Review

I made the obvious change in types but I still need to figure out the side effects.
Any help is welcomed 👋 

## To do

- [x] Update main type
- [ ] Update parser
- [ ] Make test case
- [ ] ???

### DIFF
--- a/packages/token-tools/src/types.ts
+++ b/packages/token-tools/src/types.ts
@@ -193,7 +193,7 @@ export type StrokeStyleValue =
   | StrokeStyleValueExpanded;
 
 export interface StrokeStyleValueExpanded {
-  dashArray: DimensionValue[];
+  dashArray: (DimensionValue | AliasValue)[];
   lineCap: 'round' | 'butt' | 'square';
 }
 


drwpow (approved):
This looks good to me! This is only a type change, with no impact to runtime, so it should be fine. Our test suite tests TypeScript types too and it passes, so as long as that works this should be an improvement.

On an unrelated note: I’ve been having issues with the Windows tests not completing that are unrelated to your PR. It’s either a bug with pnpm + Windows, or the runner, or something else. So for now I’m just going off the Linux tests and those are passing.

drwpow:
I know this was a draft, but I think this is complete in and of itself. In other cases we’d require tests, but I think this is safe enough to merge as-is. Thank you!

nclsndr:
> This looks good to me! This is only a type change, with no impact to runtime, so it should be fine. Our test suite tests TypeScript types too and it passes, so as long as that works this should be an improvement.

Trusting you on that one 🙌

PR #297, opened 2024-08-19
Fix Windows CI

drwpow:
## Changes

Fixes Windows test failures in CI

## How to Review

- If Windows is fixed, this worked
- If it isn’t, it didn’t


packages/cli/bin/cli.js
```diff
@@ -89,11 +93,11 @@ export default async function main() {
       const mod = await import(resolvedConfigPath);
       if (!mod.default) {
         printErrors(
-          `No default export found in ${path.relative(fileURLToPath(cwd), resolvedConfigPath)}. See https://terrazzo.dev/docs/cli for instructions.`,
+          `No default export found in ${path.relative(cwd.href, resolvedConfigPath)}. See https://terrazzo.dev/docs/cli for instructions.`,
```
drwpow, line 96: Windows users gets posix paths. It’s not worth the headache.

PR #312, opened 2024-10-14
Add Turborepo

drwpow:
## Changes

Uses [Turborepo](https://turbo.build/repo/docs) for building this repo. Not so much out of necessity; more as a learning experience for me. But also, it actually does speed things up locally

## How to Review

- CI should pass


packages/cli/test/build.test.ts
```diff
@@ -1,36 +1,9 @@
...
-import fs from 'node:fs';
 import { execa } from 'execa';
 
 const cmd = '../../../bin/cli.js';
 
 describe('tz build', () => {
-  it('default', async () => {
-    const cwd = new URL('./fixtures/build-default/', import.meta.url);
-    await execa('node', [cmd, 'build'], { cwd });
-    const testFile = new URL('./tokens/index.css', cwd);
```
drwpow, line 11: Turborepo was right for calling this out: we’re relying on a downstream package to run this test, which is backwards. Moved to plugin-css so it can be run more-easily

packages/tiles/rollup.config.js
```diff
@@ -10,25 +9,8 @@ export default {
     css({
       output: 'all-components.css',
     }),
-    ts(),
-    swc({
```
drwpow, line 14: Had a realization that we’re running `tsc` anyway to build declarations; might as well just build the React components with 1 process instead of 2 (also they won’t ever disagree; SWC was a PITA to configure and it was giving me problems again)

turbo.json
```diff
@@ -0,0 +1,17 @@
...
+      "dependsOn": ["^build"]
+    },
+    "build:app": {
+      "dependsOn": ["build"]
+    },
+    "lint": {
+      "dependsOn": ["build"]
+    },
+    "test": {
+      "dependsOn": ["build"]
```
drwpow, line 14: I actually like this—this conditionally builds for `pnpm test` but only if it needs to. Trying to find that balance using `pnpm run` has always been a struggle for me—either doing too much or too little work (and documenting it)

PR #313, opened 2024-10-15
Improve alias validation for composite types

### LINKED ISSUE #309: Failing nested aliases type validation
The tokens parser is capable of catching type mismatches for scalar types but ignore the composites.

```ts
parse([{
  src: JSON.stringify({
    color: {
      $type: 'color',
      blue: {
        $value: '#10109A',
      },
    },
    dimension: {
      base: {
        $type: 'dimension',
        $value: '{color.blue}', // should be dimension: caught by parser ✅
      },
    },
    border: {
      base: {
        $type: 'border',
        $value: '{dimension.base}', // should be border: caught by parser ✅
      },
      nested: {
        $type: 'border',
        $value: {
          color: '{color.blue}',
          width: '{color.blue}', // should be dimension: silently failing ❌
          style: 'solid',
        },
      },
    },
  })
}])
```

I think that's not much of a deal, yet I preferred to report it while looking into it.

—
The only solution I found on my end was to manually declare [the mapping](https://github.com/nclsndr/design-tokens-tools/blob/main/packages/w3c-design-tokens-parser/src/definitions/TokenTypeAliasingCompatibility.ts).

drwpow:
## Changes

Fixes #309. Adds tests and catches validation errors for aliasing composite types. Also fixes a bug in `$type: gradient` dealing with nested aliases.

## How to Review

- This is just a bugfix

### DIFF
--- a/packages/parser/parse/index.js
+++ b/packages/parser/parse/index.js
@@ -131,7 +131,12 @@ export default async function parse(
       if (mode === '.') {
         continue; // skip shadow of root value
       }
-      applyAliases(tokens[id].mode[mode], { tokens, node: tokens[id].mode[mode].source.node, logger });
+      applyAliases(tokens[id].mode[mode], {
+        tokens,
+        node: tokens[id].mode[mode].source.node,
+        logger,
+        src: sources[tokens[id].source.loc]?.src,
+      });
     }
   }
   logger.debug({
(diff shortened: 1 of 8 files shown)


packages/parser/parse/index.js
```diff
@@ -473,8 +514,20 @@ function applyAliases(token, { tokens, logger, filename, src, node }) {
             }
             const aliasOfID = resolveAlias(token.$value[i][property], { tokens, logger, filename, node, src });
             const { id: aliasID, mode: aliasMode } = parseAlias(token.$value[i][property]);
+            const aliasToken = tokens[aliasOfID];
+
+            if (expectedAliasTypes?.[property] && !expectedAliasTypes[property].includes(aliasToken.$type)) {
```
drwpow, line 519: Note: this is already a pretty large and complex function that’s doing a lot, but it’s hard to break this up because it’s handling tokens that may be invalid. In general I do like to keep logic dealing with possibly-invalid syntax centralized in one place like this, otherwise you wind up with “this may be invalid” all over your codebase.

Also because we’re in the middle of doing alias resolution which requires pretty much the whole world loaded here, passing that off to other systems probably isn’t good modularization.

All that said, this code is messy for sure, but it’s handling “the whole world” before it’s normalized and validated. The messiness, if quarantined here, helps the rest of the code remain clearer in purpose and less-concerned with edge cases.

PR #339, opened 2024-11-08
fix: add letter spacing to transform as dimension token

tomasfrancisco:
## Changes

- Added `letterSpacing` to the group of tokens to be transformed with the new dimension value/unit pair.
- Uncommented the tests for the various token types in plugin-css

## How to Review


packages/plugin-css/test/js.test.ts
```diff
@@ -11,61 +11,62 @@ import css from '../src/index.js';
...
-      const tokensJSON = new URL('./tokens.json', cwd);
-      const { tokens, sources } = await parse([{ filename: tokensJSON, src: fs.readFileSync(tokensJSON, 'utf8') }], {
-        config,
-      });
-      const result = await build(tokens, { sources, config });
-      expect(result.outputFiles.find((f) => f.filename === output)?.contents).toMatchFileSnapshot(
-        fileURLToPath(new URL('./want.css', cwd)),
-      );
-    });
+    it.each(['boolean', 'border', 'color', 'dimension', 'gradient', 'shadow', 'string', 'typography', 'transition'])(
```
drwpow, line 14: 🤦 looks like I was debugging something and accidentally committed it. thanks for re-enabling

packages/plugin-css/test/fixtures/type-transition/want.css
```diff
@@ -17,142 +17,142 @@
...
   --ds-transition-linear-quick-timing-function: var(--ds-ease-linear);
   --ds-transition-linear-moderate: var(--ds-transition-linear-moderate-duration) var(--ds-transition-linear-moderate-delay) var(--ds-transition-linear-moderate-timing-function);
   --ds-transition-linear-moderate-duration: var(--ds-timing-moderate);
-  --ds-transition-linear-moderate-delay: 0ms;
+  --ds-transition-linear-moderate-delay: 0;
   --ds-transition-linear-moderate-timing-function: var(--ds-ease-linear);
   --ds-transition-linear-deliberate: var(--ds-transition-linear-deliberate-duration) var(--ds-transition-linear-deliberate-delay) var(--ds-transition-linear-deliberate-timing-function);
   --ds-transition-linear-deliberate-duration: var(--ds-timing-deliberate);
-  --ds-transition-linear-deliberate-delay: 0ms;
+  --ds-transition-linear-deliberate-delay: 0;
```
drwpow, line 152: Hm this seems to be a regression but is probably before your PR since the tests were disabled. We can fix before shipping another release though

drwpow (approved):
This looks great, thank you!

Could you run `npx changeset` from the root of the repo and make a `patch` change to the relevant packages? That makes sure you’re credited in the changelog. Thank you!

packages/plugin-css/test/fixtures/type-transition/want.css
tomasfrancisco: This is probably because of this check in `transformDurationValue`: https://github.com/Design-System-Pro/terrazzo/blob/6a3346cc2931637e2ee9a2e0d54d58e9842a9cd5/packages/token-tools/src/css/duration.ts#L14

I actually thought was by design since this also happens in the `transformDimensionValue`: 
https://github.com/Design-System-Pro/terrazzo/blob/6a3346cc2931637e2ee9a2e0d54d58e9842a9cd5/packages/token-tools/src/css/dimension.ts#L14-L15

drwpow: Yeah you’re right. I think it was just a bad copy-paste. For some reason CSS requires units for durations always, even though dimensions may be unitless.

drwpow:
Thanks for fixing! Will cut a release later

PR #341, opened 2024-11-08
Fix duration 0 missing units

drwpow:
## Changes

Duration `0` tokens should still have a unit according to the CSS spec. This was just a bad copy-and-paste from CSS dimensions, which are allowed to be unitless at `0`.

Also bumps deps (and removes accidental monaco editor from the package lock)

## How to Review

- Tests updated


.changeset/stupid-seahorses-lie.md
```diff
@@ -0,0 +1,6 @@
+---
```
drwpow, line 1: this is the best name I have ever gotten

PR #347, opened 2024-11-13
feat(parser): output shadow inset keyword

### LINKED ISSUE #343: Inset box shadow support
Trying to setup inset shadows but couldn't find a way to add `inset` to a shadow, according to [DTCG format](https://arc.net/l/quote/ipxtuvgt). It seems though, there has been some sort of [reference](https://github.com/terrazzoapp/terrazzo/blob/be9603278f1a774233a13255c70811d0a0fe3058/packages/plugin-css/test/fixtures/type-shadow/tokens.json#L24) to it before.

Is it a missing property or is there other way we could flag an inset shadow?

tomasfrancisco:
## Changes

- Adds `inset` shadows support (Resolves #343)

### DIFF
--- a/packages/parser/parse/normalize.js
+++ b/packages/parser/parse/normalize.js
@@ -102,6 +102,7 @@ export default function normalizeValue(token) {
         offsetY: normalizeValue({ $type: 'dimension', $value: layer.offsetY ?? 0 }),
         blur: normalizeValue({ $type: 'dimension', $value: layer.blur ?? 0 }),
         spread: normalizeValue({ $type: 'dimension', $value: layer.spread ?? 0 }),
+        inset: layer.inset === true,
       }));
     }
     case 'strokeStyle': {
--- a/packages/parser/test/parse.test.ts
+++ b/packages/parser/test/parse.test.ts
@@ -1946,6 +1946,7 @@ font:
                   offsetY: { value: 0.25, unit: 'rem' },
                   blur: { value: 0.5, unit: 'rem' },
                   spread: { value: 0, unit: 'px' },
+                  inset: false,
                 },
               ],
             },
@@ -1990,6 +1991,7 @@ font:
                   offsetY: { value: 0.25, unit: 'rem' },
                   blur: { value: 0.5, unit: 'rem' },
                   spread: { value: 0, unit: 'px' },
+                  inset: false,
                 },
                 {
                   color: { colorSpace: 'srgb', channels: [0, 0, 0], alpha: 0.1 },
@@ -1997,6 +1999,7 @@ font:
                   offsetY: { value: 0.5, unit: 'rem' },
                   blur: { value: 1, unit: 'rem' },
                   spread: { value: 0, unit: 'px' },
+                  inset: false,
                 },
               ],
             },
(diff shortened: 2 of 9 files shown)


packages/parser/test/parse.test.ts
```diff
@@ -1932,6 +1932,7 @@ font:
                     offsetX: { value: 0, unit: 'rem' },
                     offsetY: { value: 0.25, unit: 'rem' },
                     blur: { value: 0.5, unit: 'rem' },
+                    inset: false,
```
tomasfrancisco, line 1935: I'm not sure if this is the best approach to make sure a `false` inset doesn't end up in the output or if is best to create an additional test.

drwpow: No I think this is right! If I remember correctly, we want to normalize/fill in all default values with the parser. Otherwise, all plugins/tools built on Terrazzo have to all remember what the defaults are, which isn’t fun for anybody.

packages/parser/parse/normalize.js
```diff
@@ -102,6 +102,7 @@ export default function normalizeValue(token) {
         offsetY: normalizeValue({ $type: 'dimension', $value: layer.offsetY ?? 0 }),
         blur: normalizeValue({ $type: 'dimension', $value: layer.blur ?? 0 }),
         spread: normalizeValue({ $type: 'dimension', $value: layer.spread ?? 0 }),
+        inset: layer.inset === true || undefined,
```
drwpow, line 105: If it doesn’t change anything, I think we should just force this to always be a boolean. That just makes things simpler down-the-line. It will default to `false`. We do the same thing with numbers/dimensions—defaulting to `0` if not specified.

```suggestion
        inset: layer.inset === true,
```

packages/token-tools/src/css/shadow.ts
```diff
@@ -28,8 +28,9 @@ export function transformShadowLayer(
   const spread = partialAliasOf?.spread
     ? transformAlias(partialAliasOf.spread)
     : transformDimensionValue(value.spread, { transformAlias });
+  const inset = value?.inset ? 'inset' : undefined;
 
-  return [offsetX, offsetY, blur, spread, color].join(' ');
+  return [inset, offsetX, offsetY, blur, spread, color].filter(Boolean).join(' ');
```
drwpow, line 33: So a subtle thing, but we actually DON’T want `.filter(Boolean)` since that may filter out any `0`s from the array, which would change the final value! Instead, we should just prepend `inset` if it’s specified. We may not have tests to catch this.

```suggestion
const value = [offsetX, offsetY, blur, spread, color];
if (inset) {
  value.unshift(inset);
}
return value.join(' ');
```

packages/token-tools/src/types.ts
```diff
@@ -466,6 +466,7 @@ export interface ShadowValueNormalized {
   offsetY: DimensionValue;
   blur: DimensionValue;
   spread: DimensionValue;
+  inset?: boolean;
```
drwpow, line 469: Let’s just enforce this in the normalized version. The intent with the normalized versions is to resolve all values, fill in defaults, and give a final, predictable shape. We can be much stricter here than the spec, because we (the parser) are the one that makes this! So we’re only helping ourselves (and plugins) out by normalizing what we can.

```suggestion
  inset: boolean;
```

drwpow (commented):
This looks fantastic, thank you! Had a couple minor things to address first before merging

packages/token-tools/src/css/shadow.ts
drwpow: Oh, wait—on second thought you’re right; at this point we should have converted all the `0`s to strings (`'0'`), so your logic would be correct! Just double-check to make sure

packages/parser/test/parse.test.ts
tomasfrancisco: Got it. I moved the "hiding" of the `inset` decision from the parser to the css plugin.

packages/token-tools/src/css/shadow.ts
tomasfrancisco: I also increased the explicit check for `true` instead of relying on typescript.

```ts
const inset = value?.inset === true ? 'inset' : undefined;
```

tomasfrancisco: And I also went with the explicit filtering. Seems more reliable and less magic, independently of what the number have been transformed to, or not.

drwpow (approved):
Perfect! Thanks so much for adding this!

PR #351, opened 2024-11-17
Fix Hue wheel

### LINKED ISSUE #336: HueWheel uniformity issue
## Description

<p align="center"></p>

The “good” don’t look so good anymore. Fix the regression in the darkening of the band. It doesn’t have to be the most saturated, but it should show the widest range of colors

## Definition of done

- [ ] Ideally the gradient operates on the OKLAB or OKLCH space. OKHSL is workable, too, if there’s a good argument to use that
- [ ] HueWheel follows [good recommendations outlined in CSS Color Module 4](https://www.w3.org/TR/css-color-4/#interpolation-space) such as alpha and color interpretation, which touch on color blending
- [ ] The ideal solution is neither too saturated no desaturated. It is one where every line has a distinct hue, and hues appear even in size.

## Notes

- Read [Two new color spaces for color picking - Okhsv and Okhsl](https://bottosson.github.io/posts/colorpicker/). This is a great summary of the trade-offs between mathematically-uniform colors vs perceptually-uniform colors vs an artistic blend between the two. We’re shooting for the artistic blend here, so some subjectivity is required

drwpow:
## Changes

Fixes Hue wheel. Caused by me playing around with something and forgetting about it. Fixes #336 

**Before**

before

**After**

after

## How to Review

- See screenshots


packages/react-color-picker/tsconfig.build.json
```diff
@@ -1,4 +1,8 @@
 {
   "extends": "./tsconfig.json",
+  "compilerOptions": {
+    "module": "NodeNext",
+    "moduleResolution": "nodenext"
```
drwpow, line 5: This is a bug with the Rollup TS plugin I’m too lazy to fix right now—it doesn’t do a good job of tracing the `extends` to resolve the final TSConfig. This is redundant, but is easier than fixing the bug upstream.

packages/react-color-picker/src/lib/webgl.ts
```diff
@@ -243,12 +227,12 @@ void main() {
   // 3 = projection toward point, hue dependent
   // 4 = adaptive Lightness, hue independent
   // 5 = adaptive Lightness, hue dependent
-  int clamp_mode = 2;
+  int clamp_mode = 3;
```
drwpow, line 230: There are quite a few changes in the WebGL code, but this was basically a fix—a one-liner. I was halfway to experimenting with some alternate methods of generating the hue wheel, but ultimately we should use the thing that looks best, and that’s changing the clamp method

packages/react-color-picker/src/lib/webgl.ts
```diff
@@ -96,17 +93,21 @@ in vec4 v_end_color;
...
 
-export class GradientRGB {
+/**
+ * Create a gradient from A to B, blended in Oklab space.
+ */
+export class GradientOklab {
   gl: WebGL2RenderingContext;
-  startColor: WebGLColor;
-  endColor: WebGLColor;
+  startColor: Oklab;
```
drwpow, line 109: Another design problem of the WebGL code was trying to make it magically handle all color spaces, all gamuts. This is far too much responsibility for it, and also isn’t necessary since at any given time, we’re only operating in one colorspace!

Instead, if we simply restrict the WebGL code to operate in the Oklab space alone, we don’t lose any performance benefits. It does what it does best—performs crazy-fast translations between Oklab <> Linear RGB. And the only time we have to calculate colorspaces on the main thread in JS is just 2 times—start and end color, which is trivial. Keeping WebGL in Oklab still reaps all the benefits but sheds a ton of complexity and potential bugs.

packages/tiles/package.json
```diff
@@ -52,9 +52,9 @@
     "@types/react-dom": "npm:types-react-dom@rc",
     "@vitejs/plugin-react-swc": "^3.7.1",
     "chokidar-cli": "^3.0.0",
-    "react": "19.0.0-rc-cae764ce-20241025",
-    "react-dom": "19.0.0-rc-cae764ce-20241025",
-    "rollup": "^4.26.0",
+    "react": "19.0.0-rc.1",
```
drwpow, line 55: RC gettin closer! 🎉

packages/cli/tsconfig.json
```diff
@@ -3,7 +3,7 @@
   "compilerOptions": {
     "baseUrl": ".",
     "module": "NodeNext",
-    "moduleResolution": "NodeNext",
+    "moduleResolution": "nodenext",
```
drwpow, line 6: VS Code would consistently autocomplete to `NodeNext` before, but today it is consistently recommending `nodenext`. Stylistic change from TS itself? Dunno.

packages/react-color-picker/src/components/ColorChannelSlider.tsx
```diff
@@ -84,18 +88,14 @@ function ColorChannelBG({ channel, color, displayMin, displayMax, min, max }: Co
...
-  if (!RGB_COLORSPACES.includes(color.original.mode)) {
-    leftColor = COLORSPACES.rec2020.converter(leftColor);
-  }
-  let rightColor = { ...color.original, [channel]: max, alpha: 1 } as WebGLColor;
-  if (!RGB_COLORSPACES.includes(color.original.mode)) {
-    rightColor = COLORSPACES.rec2020.converter(rightColor);
-  }
+  const leftColor = { ...color.original, [channel]: displayMin ?? min };
+  const rightColor = { ...color.original, [channel]: displayMax ?? max };
+  const leftOklab = useMemo(() => withAlpha(toOklab(leftColor) as Oklab) as Oklab, [color.css]);
```
drwpow, line 93: The `useColor()` hook I made because I’M sO sMArT could be better, because adjustments like this are hard (specifically, this nonsense is because the original color may be in another space, so we want to adjust the color _before_ converting, which isn’t currently possible.

I may just ditch it in favor of raw Culori methods, but for now this is a bandaid to not refactor that.

I’m honestly really happy with the performance of `useColor` as it’s designed to be used. My problem is I need more from its design 😓

PR #353, opened 2024-11-24
Lint API

drwpow:
## Changes

Re-adds Lint API, improved for Terrazzo. Adds tests, too.

## How to Review


packages/parser/src/build/index.ts
```diff
@@ -1,39 +1,26 @@
-import { isTokenMatch } from '@terrazzo/token-tools';
+import type { DocumentNode } from '@humanwhocodes/momoa';
```
drwpow, line 1: Believe it or not, moved to `.ts`. It started being _heavily_ untyped. But with all the rule additions, the scales tipped in favor of stronger, more-predictable types. It even caught a few untested bugs 🙃

PR #364, opened 2024-11-29
Fancy init

drwpow:
## Changes

clack-terrazzo

Use [Clack](https://github.com/bombshell-dev/clack) to improve the `init` prompt

## How to Review

N/A


packages/cli/test/init.test.ts
```diff
@@ -1,16 +0,0 @@
-import fs from 'node:fs';
-import { fileURLToPath } from 'node:url';
-import { execa } from 'execa';
-import { describe, expect, it } from 'vitest';
-
-describe('tz init', () => {
```
drwpow, line 6: For now, since `init` is interactive, remove the test. Perhaps we offer a `--no-interactive` version or something that’s testable, but for now, especially with a remote download, I don’t feel like bothering with flakiness of an interactive terminal.

I’m sure there’s a way to test it, but don’t want to start off with my first guess which will definitely be incorrect.

PR #392, opened 2024-12-13
Version

drwpow:
## Changes

Changesets is really not doing it for me anymore :thinking:. Having to fight with it too often recently. But not a concern for now. Only bump package versions manually because changeset failed.

## How to Review

- CI should pass


packages/plugin-js/package.json
```diff
@@ -34,7 +34,7 @@
     "test:ts": "tsc --noEmit"
   },
   "peerDependencies": {
-    "@terrazzo/cli": "^0.2.0"
+    "@terrazzo/cli": "^0.3.0"
```
drwpow, line 37: Doesn’t need a release technically because this plugin doesn’t consume any of this code. Just for the next breaking release

PR #408, opened 2025-01-16
Critical CSS fix

drwpow:
## Changes

I was today years old when I learned when the specificity increases, and CSS vars are referencing an old value, the aliases don’t cascade too 🙃. This pushes a fix that resolves a longstanding bug with plugin-css.

## How to Review

- Tests updated

PR #420, opened 2025-01-18
Fix CSS scope issue

drwpow:
## Changes

Continuation of #408 but deeper-reaching: this fixes a longstanding issue where using modes would result in stale values in CSS.

Every test and snapshot having more values is indicative of the fix. 🥳

⚠️ Minor breaking change for users of Terrazzo beta: CSS output is changed to more-closely match Cobalt 1.0. If you’ve been using Terrazzo beta this will result in different CSS vars, but in service of helping people migrate from Cobalt.

You can override this with the [variableNames](https://terrazzo.app/docs/cli/integrations/css/#config) option in plugin-css

## How to Review


package.json
```diff
@@ -28,6 +28,6 @@
     "stylelint-order": "^6.0.4",
     "turbo": "^2.3.3",
     "typescript": "^5.7.3",
-    "vitest": "^3.0.2"
+    "vitest": "^2.1.8"
```
drwpow, line 31: Downgraded back to 2.x; was seeing some uncaught processes in CI. Will update a few patches later

PR #430, opened 2025-01-29
fix(color picker): allow drag to start anywhere

lilnasy:
### Before
The drag must begin at the thumb

https://github.com/user-attachments/assets/b4dffab2-57c1-4a63-b048-7f42fe22f07c

### After

Same behavior as the native input range.

https://github.com/user-attachments/assets/26a9d8eb-d948-41d4-a545-d1bfdaa1e7ac

## How to Review

Firefox and Safari could use a second look.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -61,4 +61,4 @@ jobs:
         with:
           version: 9.15.4
           run_install: true
-      - run: pnpm test
+      - run: pnpm -r build && pnpm -r test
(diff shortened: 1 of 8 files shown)


drwpow:
This is great! Would you mind fixing the lint error? I’d be happy to merge, thank you!

The windows test is just flaky and times out (I haven’t dug into it yet)

drwpow (approved):
Thank you so much!

PR #450, opened 2025-03-13
feat(cli): embedded server for token lab

### LINKED ISSUE #371: Token Lab: `tz lab`
Create a `tz lab` CLI command that:

1. Runs the token lab locally on a server
2. Writes to `tokens.json` automatically on save (reads your local config)

This is the same lab that runs on the website, but able to update your local project.

_Note: this would mean that the Token Lab becomes a dependency of the CLI, but that’s OK! The code is already organized such that the Token Lab technically doesn’t need to depend on the CLI, just the parser._

lilnasy:
Implements and closes #371.
It took a lot of bribes to keep the tools quiet, but it works, barely but that's besides the point.

## Refactor of the token lab to allow it to be used as a component in other apps
```ts
import TokenLab from '@terrazzo/token-lab';

const tokens: string = await readFile("./tokens.json", "utf8")

<TokenLab
  tokens={tokens}
  onUpdate=((tokens: string) => { ... })
/>
```

## `tz lab`
```bash
 ~/terrazzo/packages/cli (cli-lab)
> node dist/cli.js lab
[server] Token Lab running at http://localhost:9000
```
- Vite SPA wrapper around the token lab component, served using Hono server (just the server, not the router/framework).
- Reads the token file identified by the CLI.
- Calls the onUpdate function whenever there's a change to the tokens.*

# Workarounds
- To avoid showstopping cyclci dependencies, ~~the tokens package is now compiled with the unbuilt CLI package, using node's built-in type stripping.~~ Since we need to test with Node 20 as well, that's not an option. Instead, we are using the NPM-published version of the CLI to build the tokens package.
- Removed workaround: monaco is now compiled by Vite into a worker. Previously the files were copied over from node_modules into dist.

# Future
- *Token update events, the tokens can be updated using the ui but they updated versions remain within the ui. Currently, the only way to actually update the tokens file is to edit using monaco.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,5 +29,10 @@
     "turbo": "^2.4.4",
     "typescript": "~5.7.3",
     "vitest": "^3.0.8"
+  },
+  "pnpm": {
+    "patchedDependencies": {
+      "vite-plugin-sass-dts": "patches/vite-plugin-sass-dts.patch"
+    }
   }
 }
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
     "stylelint-config-standard": "^36.0.1",
     "stylelint-order": "^6.0.4",
     "turbo": "^2.4.4",
-    "typescript": "^5.7.3",
+    "typescript": "~5.7.3",
     "vitest": "^3.0.8"
   }
 }
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "@terrazzo/monorepo",
   "private": true,
   "type": "module",
-  "packageManager": "pnpm@9.15.4",
+  "packageManager": "pnpm@10.6.2",
   "scripts": {
     "build": "turbo run build",
     "build:apps": "turbo run build:app",
--- a/packages/cli/index.html
+++ b/packages/cli/index.html
@@ -1,0 +1,12 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="UTF-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  <title>Terrazzo Token Lab</title>
+</head>
+<body>
+  <div id="app"></div>
+  <script type="module" src="./lab.tsx"></script>
+</body>
+</html> 
--- a/packages/cli/lab.tsx
+++ b/packages/cli/lab.tsx
@@ -5,4 +5,17 @@ import TokenLab from '@terrazzo/token-lab';
 const response = await fetch("/api/tokens")
 const tokenFile = await response.text()
 const root = createRoot(document.getElementById('app'));
-root.render(<TokenLab tokensFile={tokenFile} />);
+
+root.render(<TokenLab tokensFile={tokenFile} onUpdate={async updatedTokens => {
+    const response = await fetch('/api/tokens', {
+        method: 'POST',
+        headers: {
+            'Content-Type': 'text/plain'
+        },
+        body: updatedTokens
+    });
+
+    if (!response.ok) {
+        console.error(`Failed to save tokens: ${response.status}`);
+    }
+}} />);
--- a/packages/cli/lab.tsx
+++ b/packages/cli/lab.tsx
@@ -1,0 +1,8 @@
+import React from 'react';
+import { createRoot } from 'react-dom/client';
+import TokenLab from '@terrazzo/token-lab';
+
+const response = await fetch("/api/tokens")
+const tokenFile = await response.text()
+const root = createRoot(document.getElementById('app'));
+root.render(<TokenLab tokensFile={tokenFile} />);
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -41,6 +41,7 @@
   },
   "scripts": {
     "build": "tsc -p tsconfig.build.json",
+    "build:lab": "vite build",
     "dev": "tsc -p tsconfig.build.json -w",
     "format": "biome check --fix --unsafe .",
     "lint": "biome check .",
@@ -50,8 +51,10 @@
   },
   "dependencies": {
     "@clack/prompts": "^0.9.1",
+    "@hono/node-server": "^1.13.8",
     "@humanwhocodes/momoa": "^3.3.6",
     "@terrazzo/parser": "workspace:^",
+    "@terrazzo/token-lab": "workspace:^",
     "@terrazzo/token-tools": "workspace:^",
     "@types/escodegen": "^0.0.10",
     "chokidar": "^3.6.0",
@@ -60,7 +63,14 @@
     "escodegen": "^2.1.0",
     "merge-anything": "^5.1.7",
     "meriyah": "^6.0.5",
+    "mime": "^4.0.6",
     "picocolors": "^1.1.1",
     "yaml-to-momoa": "^0.0.3"
+  },
+  "devDependencies": {
+    "@vitejs/plugin-react-swc": "^3.7.2",
+    "react": "19.0.0",
+    "react-dom": "19.0.0",
+    "vite": "^6.2.1"
   }
 }
--- a/packages/cli/src/index.ts
+++ b/packages/cli/src/index.ts
@@ -1,5 +1,5 @@
-import { pathToFileURL } from 'node:url';
 import { join } from 'node:path';
+import { pathToFileURL } from 'node:url';
 import { type Config, type ConfigInit, defineConfig as defineConfigCore } from '@terrazzo/parser';
 import { cwd } from './shared.js';
 
--- a/packages/cli/src/index.ts
+++ b/packages/cli/src/index.ts
@@ -1,8 +1,7 @@
-import { createRequire } from 'node:module';
+import { pathToFileURL } from 'node:url';
+import { join } from 'node:path';
 import { type Config, type ConfigInit, defineConfig as defineConfigCore } from '@terrazzo/parser';
 import { cwd } from './shared.js';
-
-const require = createRequire(cwd);
 
 export type { Command, Flags } from './shared.js';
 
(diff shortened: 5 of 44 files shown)


packages/token-lab/src/hooks/tokens.ts
```diff
@@ -111,6 +122,22 @@ export async function loadTokens(filename: string): Promise<string> {
  * Save tokens to IndexedDB (NOT Jotai!)
  */
 export async function saveTokens(filename: string, tokens: string): Promise<void> {
+  // TODO: this logic should instead be coming from <App onUpdate={newTokens => { ... }} />
```
lilnasy, line 125: to implement the api documented when token lab was called terrazzo gallery, im guessing we will have to store the onUpdate function in a context provider. is there a cleaner approach?

packages/token-lab/src/hooks/tokens.ts
```diff
@@ -1,9 +1,11 @@
 import { type ParseResult, type TokensJSONError, defineConfig, parse } from '@terrazzo/parser';
-import radix from 'dtcg-examples/figma-sds.json';
+import radix from 'dtcg-examples/figma-sds.json' with { type: 'json' };
```
lilnasy, line 2: typescript requires json imports to be explicit.

packages/token-lab/src/hooks/navigation.ts
```diff
@@ -1,5 +1,5 @@
 import { atom, useAtom } from 'jotai';
-import { useEffect, useMemo } from 'react';
```
lilnasy, line 2: this was a type error because it was unused.

packages/token-lab/src/components/TokensNav/TokensNav.tsx
```diff
@@ -58,12 +58,12 @@ function TokensNavLevel({ path = [], data, search }: { path?: string[]; data: un
...
       // filter out non-matching tokens
       // TODO: filter out empty groups?
-      const isFiltered = search && !id.toLowerCase().includes(search);
+      const isFiltered = typeof search === 'string' && !id.toLowerCase().includes(search);
       const { $type } = parseResult.tokens[id] ?? {};
       return (
         <TreeGrid.Item key={k} id={id} hidden={isFiltered}>
           <>
-            <TokenIcon type={$type} className={c.tokenIcon} />
+            {$type !== undefined  && $type !== "string" && <TokenIcon type={$type} className={c.tokenIcon} />}
```
lilnasy, line 66: the string token type doesnt have an icon

packages/token-lab/src/components/TokensNav/TokensNav.tsx
```diff
@@ -58,12 +58,12 @@ function TokensNavLevel({ path = [], data, search }: { path?: string[]; data: un
     if ('$value' in v) {
       // filter out non-matching tokens
       // TODO: filter out empty groups?
-      const isFiltered = search && !id.toLowerCase().includes(search);
+      const isFiltered = typeof search === 'string' && !id.toLowerCase().includes(search);
```
lilnasy, line 61: type error fix

package.json
```diff
@@ -2,7 +2,7 @@
   "name": "@terrazzo/monorepo",
   "private": true,
   "type": "module",
-  "packageManager": "pnpm@9.15.4",
+  "packageManager": "pnpm@10.6.2",
```
lilnasy, line 5: 10.6 adds support for more options in pnpm-workspace. I to configure the shellemulator option to run scripts.

package.json
```diff
@@ -27,7 +27,12 @@
     "stylelint-config-standard": "^36.0.1",
     "stylelint-order": "^6.0.4",
     "turbo": "^2.4.4",
-    "typescript": "^5.7.3",
+    "typescript": "~5.7.3",
```
lilnasy, line 30: typescript 5.8 fails to find declarations for culori/fn for some reason

package.json
```diff
@@ -27,7 +27,12 @@
...
     "vitest": "^3.0.8"
+  },
+  "pnpm": {
+    "patchedDependencies": {
+      "vite-plugin-sass-dts": "patches/vite-plugin-sass-dts.patch"
+    }
```
lilnasy, line 36: the plugin emitted extensions that didnt type check in with esm projects, patched to switch from `css.d.ts` to typescript supported `d.css.ts`.

packages/cli/index.html
lilnasy, line 1: the token lab would be built by vite into `dist/lab`, the build result would be embedded into the npm package, and `tz lab` would start a static server for that folder lying inside node_modules

packages/cli/package.json
```diff
@@ -60,7 +63,14 @@
     "escodegen": "^2.1.0",
     "merge-anything": "^5.1.7",
     "meriyah": "^6.0.5",
+    "mime": "^4.0.6",
```
lilnasy, line 66: for proper  content type headers

packages/cli/src/index.ts
```diff
@@ -19,7 +18,7 @@ export function defineConfig(config: Config): ConfigInit {
         return tokenPath;
       }
       try {
-        return new URL(`file://${require.resolve(tokenPath)}`);
+        return pathToFileURL(join(process.cwd(), tokenPath));
```
lilnasy, line 21: this error'd when the token didnt start with `./`, like in the case of token-example.json in the cli package folder.

packages/cli/src/lab.ts
```diff
@@ -0,0 +1,86 @@
+import { createReadStream, createWriteStream, type PathLike } from 'node:fs';
+import { readdir } from 'node:fs/promises';
+import { fileURLToPath } from 'node:url';
+import mime from 'mime';
+import { serve } from "@hono/node-server"
```
lilnasy, line 5: its not necessary to keep hono as a dependency. i find it easier to develop with the whatwg request and response object. but now that its built, it can be easily translated into the node apis.

packages/cli/src/lab.ts
```diff
@@ -0,0 +1,86 @@
...
+  flags: Flags;
+  config: ConfigInit;
+  configPath: string;
+  logger: Logger;
+}
+
+export async function labCmd({ config, logger }: BuildOptions) {
+  /**
+   * it reads TOKEN if you squint your eyes all the way closed
+   */
```
lilnasy, line 21: outdated, i was trying 70435 but thats too large of a number to work as a port.

packages/cli/src/lab.ts
```diff
@@ -0,0 +1,86 @@
...
+          });
+        } else if (request.method === 'POST' && request.body) {
+          await request.body.pipeTo(Writable.toWeb(createWriteStream(tokenFileUrl as PathLike)));
+          return new Response(JSON.stringify({ success: true }), {
+            headers: {
+              'Content-Type': 'application/json'
+            }
+          });
+        }
+      }
```
lilnasy, line 67: /api/tokens reads and writes the tokens file

packages/cli/tokens-example.json
lilnasy, line 1: picked up a few values from the figma file for testing

packages/token-lab/package.json
lilnasy, line 1: monaco is handled entirely by vite. its simpler, and it allows packaging token lab for npm. but it makes user builds take longer, monaco is chunky.

packages/token-lab/src/components/CodeEditor/CodeEditor.tsx
lilnasy, line 1: vite statically recognises this as dynamic worker import, and builds monaco into the user project.

packages/cli/src/lab.ts
drwpow: Oh I love hono. Keep it

packages/token-lab/package.json
drwpow: Yeah as long as it’s code-split/loaded in its own chunk that’s fine. Monaco shouldn’t block pageload if we can help it (this won’t always be local there will be a hosted version where you manually save files)

But that’s a minor detail that can always be fixed later; doesn’t matter now

package.json
drwpow: Nice! Yeah this can always be updated. Turborepo just requires a version specified but we can always be on latest

packages/cli/src/index.ts
drwpow: hm we actually do need the `require.resolve` here to deal with resolving `package.json` imports correctly; some paths are rewritten. but I can take a look at it later.

packages/cli/lab.tsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import { createRoot } from 'react-dom/client';
+import TokenLab from '@terrazzo/token-lab';
+
+const response = await fetch("/api/tokens")
+const tokenFile = await response.text()
+const root = createRoot(document.getElementById('app'));
+
+root.render(<TokenLab tokensFile={tokenFile} onUpdate={async updatedTokens => {
```
drwpow, line 9: This is really clever. Love shipping a little wrapper around the lab in the CLI like this. Also with this being a Node.js tool we can use Vite’s default build without too much hassle (I usually find it lacking for packaging npm packages, but here it works great)!

packages/token-lab/index.html
```diff
@@ -3,8 +3,8 @@
   <head>
     <meta charset="utf-8" />
     <meta name="viewport" content="width=device-width" />
-    <meta name="theme-color" content="#000000" />
-    <link rel="icon" href="/favicon.svg" />
+    <meta name="theme-color" content="color(srgb .7450980392156863 .9607843137254902 0)" />
```
drwpow, line 6: 💚

drwpow (approved):
This is FANTASTIC! THANK YOUUUUUUU!

Happy to merge all this as-is without changes. This is huge!

packages/cli/test/fixtures/check-npm/terrazzo.config.js
```diff
@@ -1,5 +1,5 @@
 import { defineConfig } from '../../../dist/index.js';
 
 export default defineConfig({
-  tokens: ['my-tokens/tokens.json'],
+  tokens: ['pkg/tokens.json'],
```
drwpow, line 4: This is the only thing we’ll have to revert; it’s intentional that it’s using the name in `package.json`, NOT the relative folder name. This is needed to resolve packages in a monorepo or from npm in general. Everything else is great though!

PR #454, opened 2025-03-27
Breaking change: rename "channels" to "components"

drwpow:
## Changes

The in-progress, updated DTCG color format will use `components` rather than `channels`. Still allow both for the meantime, but provide a warning for folks for when it’s updated

## How to Review

- Tests updated
- Tests kept for both `components` and `channels` so this is somewhat backwards-compatible


packages/cli/bin/cli.js
```diff
@@ -123,7 +126,13 @@ export default async function main() {
       break;
     }
     case 'lab': {
-      await labCmd({ config, configPath, flags, logger });
+      try {
+        require.resolve('@terrazzo/token-lab');
```
drwpow, line 130: cc @lilnasy not a perfect solution, but manually requiring folks install this is a short-term solution to breaking the circular dep chain that prevents the CLI from being tested reliably, or having to pull from npm. Having to pull from published packages in the test suite could cause some regressions, and would limit us in what we’re able to test.

But once users install this I think the circular dep may still be there, so I’ll come up with another solution as this gets closer to a stable release (and we’ll celebrate and have lots of fanfare etc etc). For now, this just unblocks the CLI from releasing / auto-installing `@terrazzo/token-lab` for everyone

packages/cli/package.json
```diff
@@ -40,7 +40,7 @@
     "terrazzo": "bin/cli.js"
   },
   "scripts": {
-    "build": "tsc -p tsconfig.build.json",
+    "build": "tsc -p tsconfig.build.json && pnpm run build:lab",
```
drwpow, line 43: Test that the lab can build properly

packages/token-lab/src/app.tsx
```diff
@@ -1,4 +1,4 @@
-import '@terrazzo/tokens/dist/index.css';
```
drwpow, line 1: @lilnasy This removes one circular dep, but not all.

Between `@terrazzo/fonts`, `@terrazzo/tiles`, `@terrazzo/tokens`, `@terrazzo/react-color-picker` and `@terrazzo/use-color`, none of those packages really matter; it’s not important to me to continue publishing those and keeping them separate. The token lab is far more important, and if we wanted to, we could just shove everything React-related in here; that’d be fine. Users could even just import React components from the token-lab, too.

In the future, publishing `@terrazzo/react-color-picker` would be nice, but I’m not super concerned with it right now. Short-term I’d like to just rearrange and put things in the token lab to remove circular deps (and we don’t have to use the Terrazzo CLI in the token lab—we could just manually style everything who cares).

packages/cli/bin/cli.js
lilnasy: This is great. Until stable release, whatever allows progress is a good solution.

PR #466, opened 2025-05-07
Fix typo in JS plugin docs

drwpow:
## Changes

As reported in Discord.

Also updates deps

## How to Review

N/A


packages/parser/package.json
```diff
@@ -39,7 +39,7 @@
     "@humanwhocodes/momoa": "^3.3.8",
     "@terrazzo/token-tools": "workspace:^",
     "@types/babel__code-frame": "^7.0.6",
-    "@types/culori": "^2.1.1",
+    "@types/culori": "^4.0.0",
```
drwpow, line 42: @lilnasy I was able to update `@types/culori` to fix the type issues! (I think)

PR #470, opened 2025-05-09
feat: Tailwind v4 plugin

drwpow:
## Changes

Finally releasing the Tailwind v4 plugin (after having worked on an actual Tailwind v4 project myself).

## How to Review

- Tests added, tests should pass


drwpow:
Note: this will fail on merge to `main` because it’s unpublished. Will merge, and then manually publish

PR #477, opened 2025-05-13
feat(plugin-css): add `baseSelector` option

### LINKED ISSUE #473: `:host` selector support for css plugin
### Summary

When working with embedded apps and shadow DOM, in order to access the css variables, you need them to be declared at `:host`, not `:root` (if it's declared from within the shadow DOM).

Little demo: https://stackblitz.com/edit/stackblitz-starters-jvxz1l3d?file=index.html

It would be nice if this root selector could be customized.

As an example, Tailwindcss v4 declares all of their css variables using both selectors but from the [discussions](https://github.com/tailwindlabs/tailwindcss/discussions/15556) I see that such default behaviour (to always use `:root, :host` isn't always favorable).

### Proposal & prior art

An additional plugin option:
```
  /**
   * Set the CSS root selector, like `:root` or `:host`.
   * @default `:root`
   */
  scopeSelector?: string;
```

Maybe this needs to be more strict though?

### Extra

dzonatan:
Lets to specify the selector where CSS variables are defined.

Fixes #473


dzonatan:
should I take care of the changeset as well?

drwpow:
This looks amazing! Great tests, no notes 🙂. And yes! Please follow the comment instructions and add a `patch` changeset (while it’s 0.x, `patch` is for non-breaking feature/bugfix, even though it’s a feature it’s backwards-compatible)

dzonatan:
Got it. Interesting approach with these changesets, I was only familiar with those conventional commits.
Force-pushed in case you prefer to still have a clean git log.

drwpow (approved):
Great job, thank you! We’ll release this later today.

PR #479, opened 2025-05-14
[Draft] font shorthand and full typography token support

michaelurban:
Draft PR. No tests or guarantees included. If it looks like the right direction I'll clean it up and add tests. Updated draft version includes full typography token support in the `sass-plugin`


packages/tokens/dist/index.css
```diff
@@ -67,63 +67,94 @@
   --tz-ease-quint-in: cubic-bezier(0.64, 0, 0.78, 0);
   --tz-ease-quint-in-out: cubic-bezier(0.83, 0, 0.17, 1);
   --tz-ease-quint-out: cubic-bezier(0.22, 1, 0.36, 1);
+  --tz-font-body: var(--tz-font-body-font-weight) var(--tz-font-body-font-size)/var(--tz-font-body-line-height) var(--tz-font-body-font-family);
```
drwpow, line 70: I know it’s still in-progress, but leaving a note for now: we do need to require `font-style` to be declared, too, to set this property. Otherwise you’re blowing away other cascading token values.

I know one could argue it’s a new feature therefore it doesn’t affect _previous_ users, but I’m requiring this because it will affect _future_ users who can and will use this, and will get bitten by it. Especially with upcoming improvements to modes with the [Token Resolver spec](https://github.com/tokens-studio/resolver-spec).

drwpow: –or— alternatively, what if we did something like:

```diff
-  --tz-font-body: var(--tz-font-body-font-weight) var(--tz-font-body-font-size)/var(--tz-font-body-line-height) var(--tz-font-body-font-family);
+  --tz-font-body: var(--tz-font-body-font-style) var(--tz-font-body-font-variant) var(--tz-font-body-font-weight) var(--tz-font-body-font-size)/var(--tz-font-body-line-height) var(--tz-font-body-font-family);
   --tz-font-body-font-family: var(--tz-font-sans);
   --tz-font-body-font-size: 0.875rem;
+  --tz-font-body-font-style: inherit;
+  --tz-font-body-font-variant: inherit;
   --tz-font-body-font-variation-settings: "wght" 400;
   --tz-font-body-font-weight: 400;
   --tz-font-body-line-height: 1.4375;
```

I’d be more OK with this—this now correctly inherits from other modes and the cascade propertly. And is backwards-compatible

packages/tokens/terrazzo.config.js
```diff
@@ -1,11 +1,13 @@
 import { defineConfig } from '@terrazzo/cli';
 import { makeCSSVar } from '@terrazzo/token-tools/css';
 import css from '@terrazzo/plugin-css';
+import sass from '@terrazzo/plugin-sass';
```
drwpow, line 4: Note: this is fine for locally-testing, but all tests should be in `plugin-sass` and this package shouldn’t be modified at all. But feel free to copy examples from this into the `plugin-sass` tests just for testing

packages/tokens/dist/index.css
drwpow: Thanks for all those great thoughts! I originally thought I had clarity, but now I want to chew on all your points and think through it more holistically. I do agree that typography tokens in CSS have been a pain point, and I’d like to make it better.

One point of drift between Terrazzo and the DTCG spec is we’ve been lax about enforcing all properties are set ([test](https://github.com/terrazzoapp/terrazzo/blob/main/packages/parser/test/parse.test.ts#L2500)), whereas the spec requires all properties. The more I think about it I’m starting to agree that when setting a `typography` token, perhaps resetting all properties actually is desirable rather than deal with an unpredictable cascade of whatever-the-hell is happening in the parent scope.

I don’t know if I have a clear direction yet, just TL;DR “thinking about it.” But I think we’re both thinking about it the same way 🙂

drwpow: So to add, more clearly, **we don’t need to add more tokens that just declare default values.** In fact, that may just introduce weirdness because it’s up to browsers and user stylesheets, etc. So I think we’re sticking with your original proposal, without changes? Unless I’m missing something

drwpow:
Shipped in #589 🙂. Kept approach as-is, just made some insignificant style changes to closer match some existing code. Will still credit you in the changelog. Thank you!

PR #497, opened 2025-06-23
replace isTokenMatch with wcmatch for 5x speedup

bschlenk:
## Changes

Hi Drew! This PR removes all uses of `isTokenMatch`, instead calling `wcmatch` outside of loops and then using the returned function inside loops. I found that this dropped Figma's token build times from 35s down to 7s.

I left `isTokenMatch` since it's exported from `@terrazzo/token-tools`, but marked it as `@deprecated`. Not sure if that's the right move here, but given that it can lead to a big perf hit if used too much, I think it could be removed in the future.

Let me know if you think the `wcmatch(patterns)` call should be wrapped in a helper in `token-tools`, I could see it handling null inputs and returning a dummy function so callers don't have to handle that themselves.

---

I was profiling `tz build` and the `transform` function from wildcard-match was dominating the profile. I think it was one of those hidden quadratic complexity things because calling `wcmatch(patterns)` does a bunch of work to build up an efficient regex, but that work was being done repeatedly on the same patterns when it could be done once. 

Screenshot 2025-06-23 at 12 23 06 PM

[CPU-20250621T224815.cpuprofile](https://github.com/user-attachments/files/20868903/CPU-20250621T224815.cpuprofile)

## How to Review

This is mostly a find and replace of the `isTokenMatch` function, so it should be easy to review. The `wcmatch` function is called in the highest scope possible where the matching patterns can be compiled once by wcmatch, and then reused within loops.

### DIFF
--- a/.changeset/fine-ducks-worry.md
+++ b/.changeset/fine-ducks-worry.md
@@ -1,0 +1,12 @@
+---
+'@terrazzo/cli': minor
+'@terrazzo/parser': minor
+'@terrazzo/plugin-css': minor
+'@terrazzo/plugin-js': minor
+'@terrazzo/plugin-sass': minor
+'@terrazzo/plugin-tailwind': minor
+'@terrazzo/token-tools': minor
+---
+
+- potential 5x speedup for @terrazzo/plugin-css
+- removed isTokenMatch from @terrazzo/token-tools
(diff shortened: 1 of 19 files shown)


drwpow:
Nice! I’d be OK with just removing `isTokenMatch()` in a `minor` (breaking) change. If you could get the CI checks passing, and add a changeset (see comment), I’d be happy to merge & release 🙏

bschlenk:
I'm not sure why test-windows is failing, it doesn't seem related to my changes 🤔 it says it can't find the terrazzo.config.js file during the test

drwpow:
Oh I can take a look. Not sure either, but yeah probably unrelated. Thanks!

PR #507, opened 2025-07-02
Fix HSL colors

### LINKED ISSUE #500: terrazzo cli fails to recognize components for hsl colorSpace
### Versions

- "@terrazzo/cli": "^0.8.1",
- "@terrazzo/plugin-css": "^0.8.1"

### Node.js version

24.1.0

### OS + version

macOS 15.5

### Description

When using the HSL color space, it seems that Terrazzo CLI picks up the color space but not the component values.

Thank you, @drwpow 

### Reproduction

Here is the code as shared via Discord:

```
"$value": {
  "colorSpace": "hsl",
  "components": [218, 50, 67],
  "alpha": 1,
  "hex": "#82A1D5"
}
```

The output currently is:

```
--color-base-blue-10: hsl(none 0% 100%);
```

### Expected result

Expected output: Something like:

```
--color-base-blue-10: hsl(218deg 50% 67%);
```

### Extra

drwpow:
## Changes

Fixes #500, where HSL and HWB colors are declared `0 - 100`, rather than `0 - 1`.

Also removes the HSV colorspace, since that’s not part of CSS Color Module 4.

OKHSV is kept, only as a convenience.

## How to Review

- Tests added for all colorSpaces
- Tests should pass


packages/token-tools/src/color.ts
```diff
@@ -183,3 +190,130 @@ export function tokenToCulori(value: ColorValueNormalized): Color | undefined {
...
+    ranges: [
+      [0.0, 1.0],
+      [0.0, 1.0],
+      [0.0, 1.0],
+    ],
+  },
+  hsl: {
+    ranges: [
+      [0.0, 360.0],
+      [0.0, 100.0],
```
drwpow, line 217: The ranges here may be a little overkill. But I wanted to avoid having this hardcoded somewhere that’d be easy to miss elsewhere.

PR #513, opened 2025-07-09
fix(tailwind): use correct tailwind 4 import

camerondubas:
## Changes

The output from the tailwind plugin uses the incorrect import. This causes an error at runtime.

See https://tailwindcss.com/docs/installation/using-vite and https://tailwindcss.com/docs/upgrade-guide#removed-tailwind-directives for examples of correct usage.

## How to Review

Generate a tailwind config output and try using it

Thanks!


drwpow:
Released in 0.3.2!

PR #520, opened 2025-07-24
fix: Rework rolldown -w scripts to avoid arg parsing error on bash shell

Sidnioulz:
## Changes

Reworks `dev` commands so that `-c -w` isn't interpreted as longer a config file at `./\-w`.

## How to Review

1. Install Linux :penguin:
2. On Linux, run `pnpm dev` in a package with that script


drwpow (approved):
Oh thank you. Could also do `--watch` too but this works!

PR #530, opened 2025-08-25
Add support for JSON $refs, prepare for stable 2.0 release

### LINKED ISSUE #505: Lint rules error for missing tokens when using multiple entry points
### Versions

- `@terrazzo/cli` v0.9.0

### Node.js version

22.13.0

### OS + version

macOS 15.5

### Description

When using multiple entry points, the `a11y/min-contrast` lint rule outputs an error about the token not being found.

This is because the linter appears to operate on one entry file at a time. Therefore, if the token that needs linting is not in that file, an error will be produced, causing the entire build to fail.

### Reproduction

```js
/**
 * External dependencies
 */
import { defineConfig } from '@terrazzo/cli';
import pluginCSS from '@terrazzo/plugin-css';

export default defineConfig( {
	tokens: [
		'./tokens/color.json',
		'./tokens/font.json',
		'./tokens/radius.json',
		'./tokens/spacing.json',
	],
	outDir: './build/css',

	plugins: [
		pluginCSS( {
			filename: 'design-tokens.css',
		} ),
	],

	lint: {
		rules: {
			'a11y/min-contrast': [
				'error',
				{
					level: 'AA',
					pairs: [
						// Action pairs (ie. using step 9 as background)
						{
							foreground: 'color.primitive.primary.contrast',
							background: 'color.primitive.primary.9',
						},
						{
							foreground: 'color.primitive.primary.1',
							background: 'color.primitive.primary.9',
						},
					],
				},
			],
		},
	},
} );
```

### Expected result

No build errors.

Note that the error goes away if I comment out all entry files apart from `color.json`

### Extra

drwpow:
## Changes

Fixes #505, #405.

This is the basis for the stable version 🥳. Following this we’ll have some short testing, then the release candidate release!

- Adds support for JSON Schema `$ref`s
- Removes validation in favor of linting. This means you can **granularly opt out of validations or write your own.**
  - This isn’t a step away from validation! It’s an **upgrade.** Now you can use Terrazzo while migrating your tokens to DTCG format.
- Removes normalization, which is a big change. But with linters, you are encouraged to write tokens in the format you like.
- Rewrites a lot of the parser part
- Lots of bugfixes & docs cleanups

## How to Review

- Tests should pass


packages/parser/test/color.test.ts
```diff
@@ -0,0 +1,270 @@
...
+          },
+        ],
+        want: {
+          error: `[lint:core/valid-color] Invalid color space: undefined. Expected a98-rgb, display-p3, hsl, hwb, lab, lab-d65, lch, oklab, oklch, okhsv, prophoto-rgb, rec2020, srgb, srgb-linear, xyz-d50, xyz, or xyz-d65
+
+  3 |     "cobalt": {
+  4 |       "$type": "color",
+> 5 |       "$value": {
+    |                 ^
+  6 |         "components": [
```
drwpow, line 157: By moving validation to the linting engine, it enables much more powerful customization. Instead of getting all or nothing, folks can opt out of granular validation checks per token type, or even add their own. This introduces a lot more possibilities in how people write and manage design tokens.

packages/parser/src/parse/load.ts
```diff
@@ -0,0 +1,567 @@
...
+/**
+ * Link and reverse-link tokens in one pass.
+ *
+ * This is one of, if not the most obtuse parts of the codebase. We flip
+ * back-and-forth between DTCG dot separated IDs and JSON $ref notation. This is
+ * an “ends justifies the means” scenario where the behavior should be captured
+ * in test cases, and this just satisfies those. We’re trading off idiomatic
+ * code for performance.
+ */
+function updateFromAliasChain({
```
drwpow: This is some of the worst code I’ve ever written in my life. It’s computationally extremely efficient. But it makes no sense. I’m hoping I can think of a way to clean this up and make this more understandable before merging. But making fun of myself publicly as motivation to do so.

packages/parser/src/parse/normalize.ts
```diff
@@ -1,199 +1,147 @@
...
+  $value: unknown;
+  mode: Record<string, { $value: unknown; source: { node: any; filename: string | undefined } }>;
+}
 
-const NUMBER_WITH_UNIT_RE = /(-?\d*\.?\d+)(.*)/;
+/**
+ * Normalize token value.
+ * The reason for the “any” typing is this aligns various user-provided inputs to the type
+ */
+export function normalize(token: PreValidatedToken, { logger, src }: { logger: Logger; src: string }) {
```
drwpow, line 17: Note: move back into `normalize.ts`

packages/parser/src/parse/json.ts
```diff
@@ -209,3 +214,162 @@ export function parseJSON(input: string | Record<string, any>, options?: ParseOp
...
+    return flattened;
+  }
+  if (resolved.type !== 'Object') {
+    throw new Error(`Can’t merge $ref of type "${resolved.type}" into Object`);
+  }
+  return mergeObjects(resolved, node);
+}
+
+/** Convert valid DTCG alias to $ref */
+export function aliasToRef(alias: string): ReferenceObject | undefined {
```
drwpow: This is one of the central pieces to this: DTCG aliases are being converted into `$ref`s. Which means it will all be `$ref`s, all the way down 💯

packages/plugin-css/vitest.config.ts
```diff
@@ -8,7 +8,7 @@ export default defineConfig({
       {
         test: {
           name: 'unit',
-          include: ['test/**/.*.test.ts'],
+          include: ['test/**/*.test.ts'],
```
drwpow, line 11: Whoops accidentally turned off our CSS tests 🫠

drwpow:
As a followup, this PR does slow down performance quite a bit. But any losses are temporary and fixable; I have specific things to improve after this merges, before the stable release is cut.

PR #539, opened 2025-09-11
ci: fix Node 24 tests

drwpow:
## Changes

Fix Node 24 tests in `main`. Caused by Free GitHub Actions backing up earlier.

## How to Review

- CI should pass


.github/workflows/ci.yml
```diff
@@ -17,15 +17,15 @@ jobs:
...
-          node-version: 22
+          node-version: 24
       - uses: pnpm/action-setup@v4
         with:
           run_install: true
       - run: pnpm run lint
   test:
     strategy:
       matrix:
-        node-version: [22, 24]
```
drwpow, line 28: I’m not in the mood to deal with floating point nonsense. At least Node 24 is stable-ish across platforms. There are no other breaking changes from 22 -> 24 I’m aware of. Better to be on current version just to get better heads up

Issue #166, opened 2023-12-10
Feature: Add @property definitions to CSS plugin output


davidleger95:
`@property` definitions help enable type safety for CSS Custom Properties.

MDN Docs: https://developer.mozilla.org/en-US/docs/Web/CSS/@property

It might be good to make a flag for this in the plugin too since defining property types is optional and adds extra weight to the output.

Suggested API:

```ts
// Types
interface Options {
  // ...
  propertyDefinitions?: boolean;  // default: true
}

// Usage
pluginCSS({
  propertyDefinitions: false
})
```

deecodess:
Hey! Would love to work on this!

drwpow:
@davejsdev This actually has decent support now, so we could tackle this if we wanted to! Not as essential as some of the Token Lab stuff, but I still think it would be a great add for `plugin-css`

PR #542, opened 2025-09-12
chore: clean up indentation logic in printRule

davejsdev:
## Changes

_What does this PR change? Link to any related issue(s)._

- Clean up the nesting logic in `printRule`
- Leverage ~prettier~ Biome to format the output properly

This is mostly to prepare for addressing #166.

No functional change.

## How to Review

Read the code and run the tests. There's no functional changes, so this is mostly about code quality.


packages/plugin-css/src/lib.ts
```diff
@@ -10,6 +11,15 @@ export const FILE_PREFIX = `/* -------------------------------------------
  *  Autogenerated by ⛋ Terrazzo. DO NOT EDIT!
  * ------------------------------------------- */`;
 
+ export const formatCSS = (code: string) => {
+  // TODO: replace with Biome formatter (@biomejs/js-api)
+  return prettier.format(code, { parser: 'css' });
```
davejsdev: Added this todo since the same one exists here: https://github.com/terrazzoapp/terrazzo/blob/854cbdbefc7acbbeaf0c4c38598e6e7824193e65/packages/token-tools/src/js/index.ts#L28

packages/plugin-css/src/lib.ts
```diff
@@ -87,51 +97,87 @@ export function printRules(rules: CSSRule[]): string {
...
+      return closingBraces;
+    }
+  };
 }
 
 function _printRule(rule: CSSRule): string {
   const output: string[] = [];
   const isMediaQuery = rule.selectors.some((s) => s.startsWith('@'));
-  let indent = '';
+  const block = createCSSDeclarationBlockManager();
```
davejsdev, line 168: This abstraction removes the need to manually track the bracket nesting of blocks.

Also, rather than manually managing indentation here, we defer formatting to ~prettier~ Biome and only worry about the functional syntax here.

packages/plugin-css/test/lib.test.ts
davejsdev, line 1: Just some minor formatting fixes from prettier

packages/plugin-css/package.json
```diff
@@ -39,6 +39,7 @@
     "@terrazzo/cli": "workspace:^"
   },
   "dependencies": {
+    "@prettier/sync": "^0.6.1",
```
davejsdev: Unfortunately adding this started causing CI to fail in other packages with this strange error 🤔

```
[config] Calling `require` for "node:module" in an environment that doesn't expose the `require` function.
```

Does that make sense?

I narrowed it down to being related to this in the CLI package: https://github.com/terrazzoapp/terrazzo/blob/652c1d173838439d9197853daa254b22cfe041b4/packages/cli/src/index.ts#L15

packages/plugin-css/src/lib.ts
```diff
@@ -10,6 +11,29 @@ export const FILE_PREFIX = `/* -------------------------------------------
...
+const { projectKey } = biome.openProject();
+biome.applyConfiguration(projectKey, {
+  css: {
+    formatter: {
+      indentWidth: 2,
+      indentStyle: 'space',
+      lineWidth: 120,
+    },
+  },
+});
```
davejsdev: We could possibly let the formatter be optionally configured by the consumer in the future as well.

packages/plugin-css/package.json
davejsdev: Welp, it took me awhile but I learned how to use `createRequire()` 🙃

davejsdev:
I'll fix up the test in a bit...

drwpow:
Thanks for adding! I think this isn’t clear because of some of the tests but part of the goal of the plugin is to run in a browser. The parser and CSS & JS plugin all run in a browser and don’t require any Node.js. It’d be nice if we could find a formatting solution that meets that requirement somehow. Not that folks need this at runtime, but it is nice to be able to build playground examples using Terrazzo.

Some plugins like Sass & swift don’t run in browser, and neither does the CLI for obvious reasons. So it’s a bit patchy now.

I’m just realizing I should add tests for this via Vitest, so this is clearer.

davejsdev:
> The parser and CSS & JS plugin all run in a browser and don’t require any Node.js.

Hmm, we could use the web version of Biome. However then it can't run in node (which I assume is also important).

Could probably swap between them based on the current env though.

https://www.npmjs.com/package/@biomejs/js-api

packages/plugin-css/test/fixtures/chained-selector/want.css
```diff
@@ -42,7 +45,7 @@
   }
 }
 
-[data-color-mode="dark"][data-contrast="high"][data-product="default"], [data-color-mode="dark"][data-contrast="high"] [data-product="default"] {
+[data-color-mode="dark"][data-contrast="high"][data-product="default"],
+[data-color-mode="dark"][data-contrast="high"] [data-product="default"] {
   --color-blue-6: color(srgb 0.19215686274509805 0.5450980392156862 0.9725490196078431);
 }
-
```
drwpow, line 48: These ending newlines are important! We’ll want to keep these.

drwpow:
> > The parser and CSS & JS plugin all run in a browser and don’t require any Node.js.
> 
> Hmm, we could use the web version of Biome. However then it can't run in node (which I assume is also important).
> 
> Could probably swap between them based on the current env though.
> 
> https://www.npmjs.com/package/@biomejs/js-api

So it looks like there is an indent manager here in the formatter. What is the thing that a library is providing? Is it just line length? IMO that’s probably skippable for generated files, maybe?

For more context this cleanup is great and appreciated—reducing code complexity is important and I’m all for it. But also trying to match the user’s personal formatting settings is something I’d like to avoid if possible. Generated files should be reasonably-formatted, but most users should probably a) not commit generated files like the ones Terrazzo outputs to git, and b) if they care, they can format it post-generation using their preferred setup. Main thing is just internalizing all the complexity of formatting can be too much, in case it gets out-of-sync etc.

(I just realized in your PR that some generated `.css` is committed to git here, and that’s probably a mistake 😅 )

packages/plugin-css/test/fixtures/chained-selector/want.css
davejsdev: There are currently two ending newlines in these files, is that intentional?

This change is just removing one of them.

packages/tokens/dist/index.css
davejsdev, line 1: @drwpow I assume this is the file that isn't supposed to be committed?

davejsdev:
> So it looks like there is an indent manager here in the formatter. What is the thing that a library is providing? Is it just line length? IMO that’s probably skippable for generated files, maybe?

Biome is handling:
- Indentation (2 spaces)
- line length (but just set to the max)
- ensuring file ends with a new line
- Side effect: ensures that the text generated is valid CSS

The `createCSSDeclarationBlockManager()` doesn't do any formatting, it just handles opening and closing curly braces and keeps track of the `level` so that `closeAll()` will work without the caller needing to know how many blocks have been opened.

Your other points about formatting make sense and I agree, generated files would typically get ignored from formatters anyway if not fully gitignored like you mentioned.

My main motivation for adding Biome is so that the plugin logic can focus solely on generating valid CSS and not worry about formatting at all. Focusing on only the functional CSS simplifies the logic in printRules, but the tradeoff is that the formatting logic gets a bit more complex when separating it from the originating context, so outsourcing the formatting to Biome becomes more attractive.

---

If you're hesitant about adding the Biome dependency, that's fair. Alternatively, could expand on the `createCSSDeclarationBlockManager()` to make it handle formatting too. I feel like that might be going down a rabbit hole of building a custom CSSOM though focused on generating CSS 🤷‍♂️ (which perhaps would be a fun project to start separately)

Let me know if you want to go in a specific direction with this PR. Can also close it if you're hesitant about this solution

packages/plugin-css/test/fixtures/chained-selector/want.css
drwpow: Oh nice! Ignore me

drwpow:
> > So it looks like there is an indent manager here in the formatter. What is the thing that a library is providing? Is it just line length? IMO that’s probably skippable for generated files, maybe?
> 
> Biome is handling:
> 
> * Indentation (2 spaces)
> * line length (but just set to the max)
> * ensuring file ends with a new line
> * Side effect: ensures that the text generated is valid CSS
> …
> My main motivation for adding Biome is so that the plugin logic can focus solely on generating valid CSS 

Yeah I apologize for having hidden requirements that aren’t in tests. I try to avoid this and normally have a strategy of “if tests pass ship it.” 

Your last point of “ensuring it’s valid CSS” is really great and identified an existing hole in the current test suite. We snapshot all the CSS for every token, but you’re right there’s no guarantee those snapshots are valid CSS.

The reasons I’m hesitant about adding a formatter are mostly:

- Some formatters (Prettier) aren’t usable in browser
- Other formatters (Biome) would possibly have to bundle both Node.js and browser deps and selectively choose which (which could be tens of MBs)
- Biome also has postinstall steps that are hardware-specific

I think code-wise it would reduce unique code, but just runtime wise it could introduce some papercuts.

So to address the CSS validity question, what if instead, in our test suite where we test the generated CSS matches tokens, we ran a CSS validator on the snapshots? That way we get that very essential piece added to the tests, but in a way that doesn’t saddle the CSS plugin with a ton of weight and OS complexity. Or, if there’s a lighter-weight “pure JS” formatter/validator I wouldn’t mind adding that (specifically something that doesn’t load Node internals, or WASM bundles—if some internal Prettier package does that then I’d be OK with that as a workaround)

> Alternatively, could expand on the `createCSSDeclarationBlockManager()` to make it handle formatting too. I feel like that might be going down a rabbit hole of building a custom CSSOM though focused on generating CSS 🤷‍♂️ (which perhaps would be a fun project to start separately)

I always encourage rabbit holes! I would love to help set you up as your project if you wanted to do this! And I wouldn’t mind importing your project in here too if you wanted to make that. LMK if you ever wanted to make that because that sounds like fun for you

drwpow:
Or, alternately, we could even run Prettier/Biome formatter on our snapshots, and we merely _replicate_ some of that functionality (which we could turn off line length if we wanted to). That would also likely count as a validator and help us handle maybe the 1 or 2 additional things missing. We’d have the benefits of a formatter while still shipping portable, pure JS code that can run in a browser, Node, Deno, lambda function, etc etc

drwpow:
Just added #545 which adds browser tests. If you rebase against latest `main` then it at least should be way easier to test this.

Using any approach, if browser tests are passing, then we can weigh options! If I’m being vague I’m really not against including new packages, again, I really only care about this being able to run in browsers without _significant_ overhead (dozens of MBs and/or errors) and whatever accomplishes that is cool

davejsdev:
> I would love to help set you up as your project if you wanted to do this! And I wouldn’t mind importing your project in here too if you wanted to make that. LMK if you ever wanted to make that because that sounds like fun for you

Sure thing! :) And yes, would appreciate help getting set up. It's been awhile since I've shipped an npm package myself, so I'm not too familiar with the current tooling.

> Or, alternately, we could even run Prettier/Biome formatter on our snapshots, and we merely replicate some of that functionality (which we could turn off line length if we wanted to).

Yeah I think that's a reasonable alternative. There might be a few edge cases it doesn't fully cover since there could be some path we forget to test. So there would potentially be some benefit to emitting a runtime error/warning on invalid CSS, but gotta balance tradeoffs.

> Just added https://github.com/terrazzoapp/terrazzo/pull/545 which adds browser tests.

Perfect thanks for adding those! 🚀

packages/plugin-css/src/build/utility-css.ts
```diff
@@ -214,7 +214,7 @@ export default function generateUtilityCSS(
               output.push({
                 selectors: [selector],
                 declarations: {
-                  background: { value: `-webkit-linear-gradient(${value.value})` },
+                  background: { value: `linear-gradient(${value.value})` },
```
drwpow, line 217: Minor, but also include a note for this change in another changeset

drwpow:
Let me know if you need anything else on this PR! Yeah I just checked and `@biome/js-api` is 24MB for the web wasm bundle. We should avoid that much weight just for a formatter. Again, rebasing against `main` will have new browser tests. So any package/approach that works in-browser and is relatively lightweight is OK (ideally it’s < 100 kb but I won’t really push back strong against anything other than it being multiple MB)

davejsdev:
> Yeah I just checked and `@biome/js-api` is 24MB for the web wasm bundle.

Yeah, that's a lot. I started looking into using postcss for this. They seem to have a pretty good API for building CSS text: https://postcss.org/api/

It's a bit loose on what's accepted since they're not necessarily building pure CSS, so it would be nice to build a narrower type-safe api around it. I might just close this PR for now and explore that approach a bit more.

davejsdev:
I did some digging into [extending the CSSOM spec for generating CSS that's trustably valid](https://github.com/davejsdev/cssom-gen). While it could work, the current spec gets in the way more than I anticipated, especially for handling comment nodes. 

It seems like there'd be better ergonomics using the [PostCSS API](https://postcss.org/api/) and just wrapping it with functions that narrow down the API for our use case.

drwpow:
Cool! Yeah that sounds like a good direction.

drwpow:
I was thinking about these good suggestions again, and I’m updating plugin-css for resolvers which involves touching all this code. I think I’m going to start with [css-tree](https://www.npmjs.com/package/css-tree) which I’m not sure exactly how much weight it will add, but it would be 192 kb at the absolute max and it is tree-shakeable, so using just part of it in theory would only add a few dozen kb or so. Which I’m fine with—working with CSS is expensive. Again, as long as it’s not dozens of _megabytes_ of wasm.

drwpow:
Aaaaaaand I couldn’t use it 🙃. Turns out css-tree and cssom just… ignore comments. So in order to generate CSS with comments you have to build your own thing. Ugh.

I still ended up having some handwritten bullshit, but at least the AST types are closer to css-tree, and I added a few more tests.

PR #547, opened 2025-09-25
feat: add theme switcher to token lab

davejsdev:
## Changes

- Adds a theme switcher to Token Lab
- Adds a border to the color swatch to make colors distinct from the Token Lab background

image
image

## How to Review

1. Try out the theme switcher at the top right
2. Refresh the page to verify that theme is persisted on refresh
3. Check color swatches similar to background to see they're now distinct


davejsdev:
I noticed there's a theme switcher for the website too after adding this. My implementation is slightly different but we could probably consolidate the two eventually.

There are a few considerations for pulling this into a reusable component. I think we'd want to add props for these:
1. localstorage variable name
2. theme attribute
3. theme options (`auto` will be hardcoded because it's a special case)

packages/token-lab/src/components/EditableColorToken/EditableColorToken.module.css
```diff
@@ -11,7 +11,7 @@
 
 .swatch {
   background: var(--color);
-  border: 1px solid oklch(0 0 0 / 0.2);
+  border: var(--tz-border-3);
```
davejsdev, line 14: The border was not prominent on dark mode. I imagine we want to use theme variables anyway

packages/token-lab/src/components/Logo/Logo.tsx
davejsdev, line 1: Pulled the logo out into a component

packages/token-lab/src/components/MainNav/MainNav.tsx
```diff
@@ -1,31 +1,18 @@
 import CodePanel from '../CodePanel/CodePanel.js';
+import Logo from '../Logo/Logo.js';
+import ThemePicker from '../ThemePicker/ThemePicker.js';
 import c from './MainNav.module.css';
 
 export default function MainNav() {
   return (
     // biome-ignore lint/a11y/useAriaPropsSupportedByRole: false "banner" is nameable
     <header className={c.header} aria-label='Header'>
```
davejsdev, line 9: What is reason for having an aria-label on the header?

Doesn't this get interpreted as "header" automatically by screen readers from the tag name?

packages/token-lab/src/components/MainNav/MainNav.tsx
```diff
@@ -1,31 +1,18 @@
...
-      <h1 className={c.title}>Token Lab</h1>
-      <div className={c.codePanel}>
+      <h1 className={c.title}>
+        <Logo className={c.logo} />
+        Token Lab
+      </h1>
+      <div className={c.actions}>
+        <ThemePicker />
         <CodePanel />
       </div>
```
davejsdev, line 17: The dropdown being close to the edge of the viewport is causing a misalignment with the select content. There's inline styling from the Radix component that adds 10px margin. I think we can reduce this to either `0` or, if we want a tiny bit of margin from the viewport edge, `4px / 0.25rem`.

I didn't address this here since it's in a separate package, and it wasn't obvious if/how to override the style.

Image

packages/token-lab/src/components/MainNav/MainNav.tsx
drwpow: It’s a landmark element! And landmark elements benefit from a name. But I must have been drunk setting the aria-label to “Header” which will get announced as “Header Header.” It should be “Main” or something so it will be read as “Main. Header.” etc.

drwpow (approved):
This is all great!

Also extra context here: we had a discussion in another issue about using the Terrazzo tokens internally. I had the ambition to have more of a fully-fleshed-out design system for Terrazzo, that could be usable in other packages. But since I now work on a design system full-time for my day job, I have had less energy / concern around maintaining this. Truthfully, I don’t care about whether the Token Lab uses a custom design system or one off the shelf.

Repeating from the other thread, I _do_ care about using something where we are generating Tokens → CSS, and are able to take full advantage of all DTCG token types. I think I am 50-50 on if that’s Tailwind, but I can be convinced if you (or someone) demonstrate how to take full advantage of it (bonus points too if we ship some improvements to the Tailwind plugin).

So basically everything in `packages/tiles` can basically not be used/deleted if we don’t want to maintain it, and replaced with [whatever]. I’m OK with that. I only have the time to support core Terrazzo token generation for the next year-ish.

davejsdev:
@drwpow yeah, I think that would be the ideal end goal, but I also think that getting a functional version of the token lab is the first priority. Tailwind is great for moving quick but I find it gets in the way for stuff like pseudo elements and grid.

If I got everything in a working state using Tailwind and shadcn component, we could then get back and pull out the styles into CSS modules. That would probably be a pretty quick task for an AI agent.

I can just go off and tinker for a bit in a fork and then circle back with a demo.

What do you think?

I've got your early designs in Figma, so can work from that.

PR #550, opened 2025-09-28
feat(plugin-css): Add support for `color-scheme` property

### LINKED ISSUE #549: "color-scheme" property
### Summary

Hi!

First thanks for this set of packages. I'm starting a new design system and it really made building it so easy.

I have one minor feature request/question regarding the support of [the `color-scheme` CSS property](https://developer.mozilla.org/en-US/docs/Web/CSS/color-scheme).

There is currently no way to include that property inside modes. 

```css
:root {
    /* missing color-scheme: light */
    --color-primary: #000;
}
@media (prefers-color-scheme: dark) {
    /* missing color-scheme: dark */
    --color-primary: #fff;
}
```

### Proposal & prior art

How about adding an additional/optional `scheme` key to modeSelectors?

```js
{
    mode: 'light',
    selectors: [...],
    scheme: 'light'
}
```

As well as a `baseScheme` key for the main CSS configuration?

### Extra

Since this is a minor change, I took the freedom to submit a PR. It might be easier to see what's involved.

nlemoine:
## Changes

Provide support for the `color-scheme` CSS property by respectivley adding both `baseScheme` and `scheme` to `CSSPluginOptions` and `ModeSelector`. 

Closes #549

## How to Review

These new options shouldn't bring any breaking change as they are optional.

### DIFF
--- a/.changeset/grumpy-oranges-walk.md
+++ b/.changeset/grumpy-oranges-walk.md
@@ -1,0 +1,8 @@
+---
+"@terrazzo/plugin-css": patch
+"@terrazzo/plugin-sass": patch
+"@terrazzo/plugin-tailwind": patch
+"@terrazzo/plugin-vanilla-extract": patch
+---
+
+Add support for color-scheme property
--- a/www/src/pages/docs/cli/integrations/css.md
+++ b/www/src/pages/docs/cli/integrations/css.md
@@ -41,6 +41,7 @@ export default defineConfig({
       filename: "tokens.css",
       variableName: (id) => id.replace(/\./g, "-"),
       baseSelector: ":root",
+      baseScheme: "light dark", // Optional: support both light and dark themes
     }),
   ],
 });
@@ -309,6 +310,7 @@ export default defineConfig({
     css({
       filename: "tokens.css",
       exclude: [], // ex: ["beta.*"] will exclude all tokens in the "beta" top-level group
+      baseScheme: "light dark", // Optional: set base color-scheme
       modeSelectors: [
         {
           mode: "light",
@@ -316,6 +318,7 @@ export default defineConfig({
             "@media (prefers-color-scheme: light)",
             '[data-mode="light"]',
           ],
+          scheme: "light", // Optional: set color-scheme for this mode
         },
         {
           mode: "dark",
@@ -323,6 +326,7 @@ export default defineConfig({
             "@media (prefers-color-scheme: dark)",
             '[data-mode="dark"]',
           ],
+          scheme: "dark", // Optional: set color-scheme for this mode
         },
         { mode: "mobile", selectors: ["@media (width < 600px)"] },
         { mode: "desktop", selectors: ["@media (width >= 600px)"] },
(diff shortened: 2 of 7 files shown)


drwpow:
💯 This is fantastic, thank you!

If you could be so kind as to make a `patch` changeset, we can cut a release (even though this is a new feature, we’re using `patch` for backwards-compatible changes until we release a stable version). Be sure to add `plugin-css`, `plugin-sass`, `plugin-tailwind`, and `plugin-vanilla-extract` all to the release so they all get bumped together (since they all share output with the CSS plugin)

nlemoine:
Sure!

www/src/pages/docs/cli/integrations/css.md
```diff
@@ -431,6 +444,122 @@ The sky is the limit with mode selectors, but some popular patterns are:
 
 :::
 
+### Color Scheme
+
+The CSS plugin supports automatically generating the [`color-scheme`](https://developer.mozilla.org/en-US/docs/Web/CSS/color-scheme) property to help browsers optimize rendering for light and dark themes. This feature works with both the base selector and individual mode selectors.
```
Sidnioulz:
```suggestion
The CSS plugin supports automatically generating the [`color-scheme`](https://developer.mozilla.org/en-US/docs/Web/CSS/color-scheme) property to help browsers respect users' prefered color schemes. This feature works with both the base selector and individual mode selectors.
```

nit: I don't think theme is the right word here, and mentioning user preferences would help root what color schemes are for juniors who are still learning vocabulary.

.changeset/grumpy-oranges-walk.md
```diff
@@ -0,0 +1,8 @@
+---
+"@terrazzo/plugin-css": patch
+"@terrazzo/plugin-sass": patch
+"@terrazzo/plugin-tailwind": patch
+"@terrazzo/plugin-vanilla-extract": patch
```
Sidnioulz, line 5: Are you sure that these three packages are impacted?

www/src/pages/docs/cli/integrations/css.md
```diff
@@ -309,20 +310,23 @@ export default defineConfig({
...
           ],
+          scheme: "light", // Optional: set color-scheme for this mode
         },
         {
           mode: "dark",
           selectors: [
             "@media (prefers-color-scheme: dark)",
             '[data-mode="dark"]',
           ],
+          scheme: "dark", // Optional: set color-scheme for this mode
```
Sidnioulz, line 329: Question: what happens if the baseScheme is set to e.g. `only light` and scheme `dark` is passed? Or in other words, if an invalid scheme is passed?

Would it be useful to write some config validation?

www/src/pages/docs/cli/integrations/css.md
nlemoine: Salut Steve, good point, agreed!

.changeset/grumpy-oranges-walk.md
nlemoine: That's what @drwpow [told me](https://github.com/terrazzoapp/terrazzo/pull/550#issuecomment-3347324318) 🤷‍♂️

www/src/pages/docs/cli/integrations/css.md
nlemoine: IMO, I would leave this to user responsibility (and the valid properties might change over time which brings maintenance burden).

I don't know the library well enough yet to catch the spirit, are config values validated elsewhere?

From a quick test in the `tokens` package, it doesn't look like there's any validation:
```js
// in modeSelectors
{
  mode: 'light',
  selectors: ['hello world', '[data-color-mode="light"][data-color-mode="light"]'],
},
```

Outputs:
```css
hello world,
[data-color-mode="light"][data-color-mode="light"] {
  --tz-border-1: 1px solid var(--tz-color-border-1);
```

That's totally ok and out of this library's scope to me. 

But I'll let you evaluate if it's worth validating or not.

.changeset/grumpy-oranges-walk.md
Sidnioulz: Oh that makes sense, thanks for sharing.

www/src/pages/docs/cli/integrations/css.md
Sidnioulz: I don't have an opinion myself, and just figured I'd raise the question to see what you thought.

You have a solid reasoning informing your choice, I agree with you wrt the maintenance burden. I'm just not sure how hard it would be for users to debug a human error in setting the config without validation Vs having immediate error feedback.

drwpow (approved):

PR #552, opened 2025-09-29
fix(parser): Apply transform to token nodes

nlemoine:
## Changes

The parser was failing to apply transforms when $type was declared at the token level. The code only checked for inheritedTypeNode (from group-level inheritance), but didn't check for token-level $type declarations.

## How to Review

Should be non breaking, transform are still applied at group level.

However, people using the API might have now to deal with token level transforms.


nlemoine:
@drwpow Sorry for the PR flood, I started using many `@terrazzo` packages and found some tiny bugs 😄

Sidnioulz (approved):
I can confirm the test fails without this code, and it also fails on my branch which has slightly reworked inherited type computation.

Token `$type` fields are ignored in `parse`, and the inherited and local types are only consolidated in `validateTokenNode` to produce a well-defined token.

@nlemoine if anything I'd just change the JSDoc comment a little bit to clarify that the change made affects `transform` specifically.

packages/parser/src/parse/index.ts
```diff
@@ -256,8 +256,18 @@ async function parseSingle(
         const inheritedTypeNode = computeInheritedProperty(node, '$type', { subpath, inherited: $typeInheritance });
         if (node.value.type === 'Object') {
           const $value = node.value.members.find((m) => m.name.type === 'String' && m.name.value === '$value');
-          if ($value && inheritedTypeNode?.value.type === 'String' && transform?.[inheritedTypeNode.value.value]) {
-            const result = transform[inheritedTypeNode.value.value]?.(evaluate(node.value), subpath.join('.'), node);
+
+          // Check for both inherited type and token-level type
```
Sidnioulz:
```suggestion
          // Call transform using either inherited type or token-level type
```

Just to clarify that this only impacts parse/transform and not the content of tokens once they are parsed.

nlemoine: 👍 Thanks for the comment improvement, makes more sense indeed

drwpow (approved):
Keep the PRs coming, thank yoU!

PR #553, opened 2025-09-30
Introduce the Token Listing plugin

Sidnioulz:
## Changes

I added the token listing plugin :)

## How to Review

* Run tests for the plugin
* Build your own token listing for your fav design system and provide feedback :)

## What I expect to be wrong

I'm unhappy with how `originalValue` works. I think it's less useful in the listing than the `aliasChain` you use internally and I intend to review the RFC accordingly.

I have not updated the JSON schema and RFC in some time. I'll do that on Thursday.


.vscode/settings.json
```diff
@@ -22,5 +22,6 @@
   },
   "[typescriptreact]": {
     "editor.defaultFormatter": "biomejs.biome"
-  }
+  },
+  "vitest.disableWorkspaceWarning": true
```
Sidnioulz, line 26: I've set that because Vitest is freaking out a bit in this monorepo and missing quite a few tests, but my feeling is that most test suites are very small and it's acceptable to scan multiple projects in this instance.

.changeset/full-webs-shake.md
```diff
@@ -0,0 +1,11 @@
+---
+"@terrazzo/plugin-vanilla-extract": minor
+"@terrazzo/plugin-token-listing": minor
+"@terrazzo/plugin-swift": minor
+"@terrazzo/plugin-sass": minor
+"@terrazzo/plugin-css": minor
+"@terrazzo/plugin-js": minor
+"@terrazzo/parser": minor
```
Sidnioulz: Most of these plugins receive token listing support, hence the bump. Only Tailwind lags behind.

packages/parser/src/build/index.ts
```diff
@@ -192,7 +194,13 @@ export default async function build(
   const startBuildEnd = performance.now();
   for (const plugin of config.plugins) {
     if (typeof plugin.buildEnd === 'function') {
-      await plugin.buildEnd({ outputFiles: structuredClone(result.outputFiles) });
+      await plugin.buildEnd({
+        context: { logger },
+        tokens,
+        getTransforms,
+        sources,
+        outputFiles: structuredClone(result.outputFiles),
```
Sidnioulz, line 194: There's an unresolved type error here as `build` uses `OutputFile` but `buildEnd` expects `OutputFileExpanded`.

.changeset/full-webs-shake.md
Sidnioulz: - [ ] fix majors, they're accidents

www/src/pages/docs/cli/integrations/token-listing.md
```diff
@@ -0,0 +1,525 @@
...
+* documentation websites
+* MCP servers
+* linting plugins
+* code migration tools
+* search indexes for design tokens
+
+
+Below are a few screenshots of what such tools can look like:
+
+// TODO
```
Sidnioulz: - [ ] Missing the screenshots

packages/plugin-token-listing/test/cli.test.ts
```diff
@@ -0,0 +1,74 @@
...
+  describe('plugin-token-listing options', () => {
+    it('outDir', async () => {
+      const cwd = new URL('./fixtures/cli-config-outdir/', import.meta.url);
+      await execa('node', [cmd, 'build'], { cwd });
+      
+      const actual = fs.readFileSync(new URL('./styles/out/actual.listing.json', cwd), 'utf8');
+      const comparable = actual.replace(/file:\/\/\/[^"]+/g, 'file:///some-path/tokens.json');
+      await expect(comparable).toMatchFileSnapshot(
+        fileURLToPath(new URL('./styles/out/want.listing.json', cwd)),
+      );
```
Sidnioulz: * [x] Somehow windows reports different file offsets in source, also mock that

.changeset/full-webs-shake.md
drwpow: If these are all non-breaking changes, this should be a `patch` release. But I’m not sure it matters; I do have some semi-breaking changes coming up soon anyway (and we’ll release this on a beta either way)

.vscode/settings.json
drwpow: Yup this is fine

packages/parser/src/build/index.ts
```diff
@@ -87,6 +86,7 @@ export default async function build(
   for (const plugin of config.plugins) {
     if (typeof plugin.transform === 'function') {
       await plugin.transform({
+        context: { logger },
```
drwpow, line 89: Completely aside, one open question I’ve had is that `logger.error()` throws an error. But given that it’s configurable, TypeScript doesn’t treat code that follows as dead code. Perhaps the only way to fix it is just separate `error()` from `errorWithoutThrow()` or something, rather than the `{ continueOnError: boolean }` option. AFAIK TypeScript does do a good job of “remembering” when a function throws

www/src/pages/docs/cli/integrations/token-listing.md
```diff
@@ -0,0 +1,525 @@
...
+
+The `ModeOption` object contains the following properties:
+* `name`: the name of the mode
+* `values`: possible values for the mode
+* `description`: an optional human-readable description of the mode
+* `default`: an optional default value
+
+### PlatformOption
+
+To reuse an existing Terrazzo plugin that's compatible with the Token Listing plugin, pass the name of the plugin. Check your token listing output for token names to verify compatibility.
```
pvignau, line 533: It'd be interresting to update it and focus that it is the transform name that is used here and not directly the plugin. 

What if we have several `Transforms` ? 🤔 Need to use a `function` ?

packages/plugin-token-listing/src/build.ts
```diff
@@ -0,0 +1,170 @@
...
+  }
+
+  let filter: boolean = true;
+  if ('filter' in platform && typeof platform.filter === 'string') {
+    filter = !!getNameFromPlugin({ getTransforms, mode, plugin: platform.filter, token });
+  } else if ('filter' in platform && typeof platform.filter === 'function') {
+    filter = platform.filter({ logger, mode, tokensSet, token });
+  }
+
+  return filter ? name : undefined;
```
pvignau, line 83: How about throw an error in case a platform is defined but can not be found while building ? Or at least a warning ?

pvignau (commented):
Nice Work ! Works like a charm. 

I'd have some changes to suggest, not on the code but the feature. 

We discussed about the absolute filepath on the `source.resource` key. It'd be really great to have a relative path or maybe a configuration key to force this attr. 

Also, I have a question for mode handling, I don't know what to think about but let's take these examples : 
```json
{
      "$name": "color.bg.accent.strong.enabled",
      "$type": "color",
      "$value": {
        "colorSpace": "srgb",
        "components": [
          0,
          0.47843137254901963,
          0.592156862745098
        ],
        "alpha": 1,
        "hex": "#007a97"
      },
      "$extensions": {
        "app.terrazzo.listing": {
          "names": {
            "scss": "--color-bg-accent-strong-enabled"
          },
          "originalValue": "{core.color.blue.500}",
          "previewValue": "#007a97",
          "source": {
            "resource": "file:///Users/pyvignau/workspace/ftven/galaxie/packages/new-tokens/src/dtcg/tokens.json",
            "loc": {
              "start": {
                "line": 76,
                "column": 22,
                "offset": 2474
              },
              "end": {
                "line": 85,
                "column": 12,
                "offset": 2764
              }
            }
          }
        }
      }
    },
    {
      "$name": "color.bg.accent.strong.enabled",
      "$type": "color",
      "$value": {
        "colorSpace": "srgb",
        "components": [
          0,
          0.47843137254901963,
          0.592156862745098
        ],
        "alpha": 1,
        "hex": "#007a97"
      },
      "$extensions": {
        "app.terrazzo.listing": {
          "names": {
            "scss": "--color-bg-accent-strong-enabled"
          },
          "originalValue": "{core.color.blue.500}",
          "previewValue": "#007a97",
          "mode": "light",
          "source": {
            "resource": "file:///Users/pyvignau/workspace/ftven/galaxie/packages/new-tokens/src/dtcg/tokens.json",
            "loc": {
              "start": {
                "line": 76,
                "column": 22,
                "offset": 2474
              },
              "end": {
                "line": 85,
                "column": 12,
                "offset": 2764
              }
            }
          }
        }
      }
    },
    {
      "$name": "color.bg.accent.strong.enabled",
      "$type": "color",
      "$value": {
        "colorSpace": "srgb",
        "components": [
          1,
          1,
          1
        ],
        "alpha": 1,
        "hex": "#ffffff"
      },
      "$extensions": {
        "app.terrazzo.listing": {
          "names": {
            "scss": "--color-bg-accent-strong-enabled"
          },
          "originalValue": "{core.color.blue.500}",
          "previewValue": "#ffffff",
          "mode": "dark",
          "source": {
            "resource": "file:///Users/pyvignau/workspace/ftven/galaxie/packages/new-tokens/src/dtcg/tokens.json",
            "loc": {
              "start": {
                "line": 76,
                "column": 22,
                "offset": 2474
              },
              "end": {
                "line": 85,
                "column": 12,
                "offset": 2764
              }
            }
          }
        }
      }
    },
```

These are two tokens with different modes `light` and `dark`, `previewValue` is correct but the original value is the same on both (uses the `light` mode value)
This is not a false value as long as the absolute default value for me is the light one but the original value for the `dark` mode is wrong where `blue.500` is not `#ffffff`.

Cheers !

.changeset/full-webs-shake.md
Sidnioulz: I was aiming for minor as these plugins change to support token listing. So I see that more as a feature than a bug fix.

I am however baffled that I received an email from Changesets-bot telling me two plugins would release under a major version, when clearly I've only set minors :sweat_smile: 

image

packages/parser/src/build/index.ts
Sidnioulz: I've ended up adjusting output file types, I think it was just the wrong type being applied as the properties that were missing from the `buildEnd` are indeed computed by the CLI.

packages/parser/src/build/index.ts
Sidnioulz: I think you can do that but you'd probably need to massage the input types a bit.

```ts
/** Log an error message (always; can’t be silenced) */
  error(entry: LogEntry & { continueOnError: true }): undefined;
  error(entry: LogEntry & { continueOnError: false | undefined }): never;

  error(entry: LogEntry) {
    this.errorCount++;
    const message = formatMessage(entry, 'error');
    if (entry.continueOnError) {
      // biome-ignore lint/suspicious/noConsole: this is a logger
      console.error(message);
      return;
    }
    if (entry.node) {
      throw new TokensJSONError(message);
    } else {
      throw new Error(message);
    }
  }
```

And then this shows dead code after returning:
```ts
const moreSpecificEntry: LogEntry & { continueOnError: false } = {
      ...entry,
      continueOnError: false,
    };
    this.error(moreSpecificEntry);
```



This would work more easily:
```ts
error(entry: LogEntry, continueOnError: true): undefined;
  error(entry: LogEntry, continueOnError: false): never;

  error(entry: LogEntry, continueOnError: false | true) {
    ...
  }
```

Sidnioulz: Throwing means reaching a non-satisfiable type for a type inference system, which TS supports through `never`. And it also supports signature overrides on functions, so the only difficulty is ensuring that param types can always properly be inferred to a single signature at check time rather than runtime. `continueOnError` must always be statically hardcoded and never be a variable.

Sidnioulz:
> This all looks great! I think I’d still like to see a couple tests, even if they’re very simple smoke tests. But once those are in, would be happy to merge & release.

Do you have specific things in mind? I'm at 84% branch coverage at the moment and not sure in which direction to take things to make tests better.

.changeset/full-webs-shake.md
drwpow-figma: > So I see that more as a feature than a bug fix.

This is true for libraries that are publishing > 1.0! But it’s expected for 0.x to play by different rules since Major versions aren’t possible. And up till now we’ve followed the pattern pretty closely that all patches are backwards-compatible and minors are the only versions with breaking changes

However, I did merge #530 which is a major update and had some breaking changes, so it’ll be a minor regardless :P

> I am however baffled that I received an email from Changesets-bot telling me two plugins would release under a major version,

This is hands down the biggest annoyance I have with Changesets—their version bumping algorithm is just wrong. They explain why they do it, and how it works, but I still have to override it all the time 🙃. I’d probably fork Changesets and/or build my own if I had the interest to build this myself, but I don’t, so I just put up with it

Sidnioulz:
> We discussed the absolute filepath on the `source.resource` key. It'd be really great to have a relative path or maybe a configuration key to force this attr.
> 
> ```json
>     {
>       "$name": "color.bg.accent.strong.enabled",
>       /* ... */
>       "$extensions": {
>         "app.terrazzo.listing": {
>           /* ... */
>           "source": {
>             "resource": "file:///Users/pyvignau/workspace/ftven/galaxie/packages/new-tokens/src/dtcg/tokens.json",
>             "loc": {
>               "start": { "line": 76, "column": 22, "offset": 2474 },
>               "end": { "line": 85, "column": 12, "offset": 2764 }
>             }
>           }
>         }
>       }
>     },
> ```

@drwpow how do you think the Resolver Spec will affect that?

Right now, to provide source info on a token, I copy your own internal source object. It's great as it points to the correct resource and location, but it's absolutely resolved.

Instead, I want it to be resolved relatively so the `resolver.json` config. If `resolver.json` points to external files, I want the same exact file paths:
```jsonc
>             "resource": "path/to/resource.json",
>       // or "resource": "file://<root>/path/to/resource.json",
>             "loc": { /* ... */ }             }
```

If it has inline tokens, I want to point to the resolver file itself:
```jsonc
>             "resource": "file://<root>/resolver.json",
>             "loc": { /* ... */ }             }
```

I'm also not entirely sure whether Token Listing should embed resolver configs, because of the risk posed by large inline token sets. WDYT?

drwpow-figma:
> > This all looks great! I think I’d still like to see a couple tests, even if they’re very simple smoke tests. But once those are in, would be happy to merge & release.
> 
> Do you have specific things in mind? I'm at 84% branch coverage at the moment and not sure in which direction to take things to make tests better.

Oh, apologies—for some reason when I looked at it, the test file didn’t load (I think I was using the new GitHub UI which I’ve had problems with files not loading at times). I do see your test and it looks great.

drwpow-figma:
> @drwpow how do you think the Resolver Spec will affect that?
>
> Right now, to provide source info on a token, I copy your own internal source object. It's great as it points to the correct resource and location, but it's absolutely resolved.
>
> Instead, I want it to be resolved relatively so the resolver.json config. If resolver.json points to external files, I want the same exact file paths:
>             "resource": "path/to/resource.json",
>       // or "resource": "file://<root>/path/to/resource.json",
>             "loc": { /* ... */ }             }
> If it has inline tokens, I want to point to the resolver file itself:

Hm we could probably chat about this sync, but in general, one change we’re incorporating from the RFC is the request that the resolver spec _MAY_ reference tokens/sets within the same file. So to accommodate that, we’ll also be including support for JSON `$ref` ([example](https://redocly.com/learn/openapi/ref-guide)). Specifically, we’ll allow folks to use JSON Schema `$defs`, and then reference them via `$ref`.

When “bundling” all these things, some resolver specs may choose to have a single file, and use `$ref`. Some may choose to still point to remote files. But either way it won’t be a hard requirement for everything to live apart from the resolver file itself.

From the standpoint of Terrazzo, though, we still do resolve things absolutely in the end because that’s how we track circular and/or invalid `$ref`s. So the `URL` behavior like we’re doing will probably stay. But also, if you choose to express things in terms of `$ref` syntax, that will work too, more than likely (again, since #530 merged)

drwpow (approved):

Sidnioulz:
> These are two tokens with different modes `light` and `dark`, `previewValue` is correct but the original value is the same on both (uses the `light` mode value)
> This is not a false value as long as the absolute default value for me is the light one but the original value for the `dark` mode is wrong where `blue.500` is not `#ffffff`.

That was a bug, thanks for catching it. I'm in the process of fixing it!

.changeset/full-webs-shake.md
Sidnioulz: > This is true for libraries that are publishing > 1.0! But it’s expected for 0.x to play by different rules since Major versions aren’t possible. And up till now we’ve followed the pattern pretty closely that all patches are backwards-compatible and minors are the only versions with breaking changes

I have a deeply-rooted grudge against semver tools that don't respect < 1.0 conventions :angry: Why do we have to account for that instead of the tool? :anger:

* [x] I see everyone is in 0.x, so I'll make all changes `patch` on my PR :)

> This is hands down the biggest annoyance I have with Changesets—their version bumping algorithm is just wrong. They explain why they do it, and how it works, but I still have to override it all the time 🙃. I’d probably fork Changesets and/or build my own if I had the interest to build this myself, but I don’t, so I just put up with it

I feel you. I wish there was _one_ tool that just really implemented semver through and through between s-r, auto, changesets... I got so mad at semantic-release's lack of monorepo support 2 years ago that I built @pvignau and I a tailored semver algorithm that tracks our dependencies :clown_face: I won't be caught having to maintain such a monster in a FOSS project though!

Sidnioulz:
> Oh, apologies—for some reason when I looked at it, the test file didn’t load (I think I was using the new GitHub UI which I’ve had problems with files not loading at times). I do see your test and it looks great.

All good, I did have part of the format not tested, and believe it or not, there was a bug exactly there. :upside_down_face:

Sidnioulz:
> From the standpoint of Terrazzo, though, we still do resolve things absolutely in the end because that’s how we track circular and/or invalid `$ref`s. So the `URL` behavior like we’re doing will probably stay. But also, if you choose to express things in terms of `$ref` syntax, that will work too, more than likely (again, since #530 merged)

Hm, makes sense. Yeah, I'd like to steal some of your time to talk about that sync. I do want to ensure that I output something as DTCGish as possible, so that I have better interop out of the box with other tools and data hosting contexts.

packages/plugin-token-listing/src/build.ts
Sidnioulz: * [x] Add UTs to cover non-tz naming patterns
* [x] logger.error if platform name is passed but no transform found
* [x] drastically improve www docs to clarify requirement

packages/plugin-token-listing/src/build.ts
```diff
@@ -0,0 +1,174 @@
...
+
+    const source = token.source.loc
+      ? {
+          resource: token.source.loc,
+          loc: token.source.node.loc,
+        }
+      : undefined;
+    if (source) {
+      output.source = source;
+    }
```
Sidnioulz: * [x] Provide config endpoint or build logic to replace local absolute path with `<root>`
* [x] Add UTs
* [x] Clarify in docs that Resolver Spec will influence this based on chat with Drew

packages/plugin-token-listing/src/build.ts
Sidnioulz: Should be better now, feel free to reopen the conversation if you still have issues :)

packages/plugin-token-listing/src/build.ts
Sidnioulz: Almost done with that. Wanna circle back with Drew but I've built something a lot more stable in the meantime. I expect there'll be a v2 post Resolver Spec anyway so I'm ok with releasing in the current state. LMK what you think.

packages/plugin-token-listing/src/build.ts
```diff
@@ -0,0 +1,193 @@
...
+
+    return output;
+  };
+
+  return async function build({ context: { logger }, getTransforms, outputFile, tokens }) {
+    const listing = Object.values(tokens).flatMap((token) =>
+      Object.entries(token.mode).map(([mode, tokenInMode]) => ({
+        $name: token.id,
+        $type: token.$type,
```
Sidnioulz, line 179: * [x] $description is missing

www/src/pages/docs/cli/integrations/token-listing.md
Sidnioulz, line 1: * [x] $description missing in the example output

www/src/pages/docs/cli/integrations/token-listing.md
Sidnioulz: Not gonna include any for now to keep things simple and employer-conversation-free

www/src/pages/docs/cli/integrations/token-listing.md
Sidnioulz: If you have several transforms, you can target the format id of either transform you want.

I'll try and rework that further when I turn my attention to the Tailwind plugin (plugin-driven subtyping + multiple name outputs on the same plugin).

Sidnioulz: I also realise I must allow multiple platform values to be stored in the listing... :sleeping:

Sidnioulz: > I also realise I must allow multiple platform values to be stored in the listing... 😴

This will be handled in an update prior to the official v1 as I won't have time to handle it here. Adding a notice in the docs that the implem is WIP

Sidnioulz:
@drwpow I can't check the Cloudflare Pages logs so can't find what the error is. Builds do pass locally.

Also unsure what to do with the Windows UT failures. It seems I can no longer use `path.relative` so I don't know how to properly resolve those `source` paths I'm retrieving from Terrazzo. Any ideas?

packages/react-color-picker/biome.json
```diff
@@ -1,4 +1,5 @@
 {
+  "root": false,
```
drwpow, line 2: This is optional when `extends: "//"` exists [according to their documentation](https://biomejs.dev/guides/big-projects/#monorepo). We can delete this!

www/src/components/docs-sidenav.astro
```diff
@@ -2,7 +2,7 @@
 import { getAriaCurrent } from '../lib/navigation.js';
 import Terrazzo from './terrazzo.astro';
 
-const NAV_ITEMS = [
+const _NAV_ITEMS = [
```
drwpow, line 5: Why this change?

packages/plugin-token-listing/.npmignore
```diff
@@ -0,0 +1,7 @@
+.turbo
+biome.*
+rolldown.*
+src/**
+test/**
```
drwpow, line 5: There’s no `test` folder in this package. So this will ship all `*.test.*` files to npm. IMO we shouldn’t ship test files to npm, so moving them in there is preferable.

Either that, or change this `.npmignore` file to also ignore `**/*.test.*`, but then it’s inconsistent with the rest

packages/plugin-token-listing/test/cli.test.ts
```diff
@@ -0,0 +1,76 @@
...
+  describe('plugin-token-listing options', () => {
+    it('outDir', async () => {
+      const cwd = new URL('./fixtures/cli-config-outdir/', import.meta.url);
+      await execa('node', [cmd, 'build'], { cwd });
+
+      const actual = fs.readFileSync(new URL('./styles/out/actual.listing.json', cwd), 'utf8');
+      const comparable = actual
+        .replace(/"line": \d+/g, '"line": 0')
+        .replace(/"column": \d+/g, '"column": 0')
+        .replace(/"offset": \d+/g, '"offset": 0');
```
drwpow, line 72: Suggestion: for Windows, IMO it is good to test the snapshot. So we could do something that just checks for e.g.

```ts
.replace(/"resource": "file:\/\/.*\/build-default\/tokens\.json", '"resource": "file://<root>/build-default/tokens.json"')
```

www/src/components/docs-sidenav.astro
Sidnioulz: I ran `biome` on the whole monorepo and it moved. So I figured I might as well embrace the linter's preference. Happy to revert that commit if you prefer though.

packages/react-color-picker/biome.json
Sidnioulz: If there is no harm in it, having it would avoid having to re-remove it from all packages when running `biome migrate`. Here as well, happy to revert!

packages/plugin-token-listing/.npmignore
Sidnioulz: I'm confused. I think I might be too tired to understand what you're saying. What should I change?

For the record I copied this from `plugin-css`, so any adjustment made here might need to be made on many/all other packages.

packages/plugin-token-listing/test/cli.test.ts
Sidnioulz: Thanks for the suggestion!

I'll first rework the code to use URLs because it's what I should have done, and if I'm lucky, it might fix the issue.

I really really want those URLs to be stable across build environments until I can make something more in line with resource names in `resolver.json`.

drwpow: I hear you but “stable across all build environments” is very tricky! There’s not only POSIX vs Windows, this may also have different behavior in containers as well as tools like Bazel. It’s generally rare for tools that deal with resolution to be platform-agnostic; even bundlers don’t do this. Anyways, if the “juice is worth the squeeze” go for it, but there should be clear advantages

PR #557, opened 2025-10-04
Speed up parser

drwpow:
## Changes

After #530, add some minor-but-surprisingly-noticeable speedups. These fixes are minor, mostly shuffling around `.sort()`s to only be called the minimum number of times. And reusing some arrays in memory which do make a difference for large token sets. It’s not perfect, but I’m seeing about ~10%-15% faster speedups in the parser.

## How to Review

- Not captured in tests. Benchmark tests are kinda cursed. Just trust me bro


packages/parser/src/parse/load.ts
```diff
@@ -63,7 +63,6 @@ export async function loadSources(
 
   // 1. Bundle root documents together
   const firstLoad = performance.now();
-  logger.debug({ ...entry, message: `Loading token files` });
```
drwpow, line 66: Unrelated: we don’t need as many subtasks to announce when they start.

packages/parser/src/parse/load.ts
```diff
@@ -162,40 +164,41 @@ export async function loadSources(
   }
 
   // 2c. DTCG alias resolution
-  // Note: these have to be resolved after $refs. Because they technically
-  // aren’t convertible to $refs until the final document is built.
+  // Unlike $refs which can be resolved as we go, these can’t happen until the final, flattened set
   resolveAliases(tokens, { logger, sources: sourceByFilename, refMap });
-  logger.debug({ ...entry, message: 'Parsing done', timing: performance.now() - resolveStart });
+  logger.debug({ ...entry, message: 'Parsing: 2nd pass', timing: performance.now() - secondPass });
```
drwpow, line 169: Added a few more minor debugging points to see which areas are the bottlenecks.

Reassuringly, a lot of the time spent is on the hardest work, like actually parsing the tokens. We’re also not paying a ton of “cost” for multiple files, too!

I did profile 2 major areas that still need deeper investigation: `tokenFromNode()` is still the bulk of the time. As is the `transform()` hook in plugins. The 2 points when things are being shoved into memory. Will dig more into both of those in followups—I’m sure there are more improvements there.

---

To my surprise, I thought putting work on the linting API would be the slowdown. But that’s actually really really fast, even with every lint rule re-scanning tokens.

PR #586, opened 2025-11-15
feat: Place `color-scheme` before CSS properties

nlemoine:
## Changes

Not much, it just place the `color-scheme` before CSS properties.

Although there's nothing wrong being at the end, it's usually used on top of CSS props. It also aligns better with the doc examples. 

## How to Review

Tiny change, nothing to worry about.

### DIFF
--- a/.changeset/modern-foxes-wave.md
+++ b/.changeset/modern-foxes-wave.md
@@ -1,0 +1,8 @@
+---
+"@terrazzo/plugin-css": patch
+"@terrazzo/plugin-sass": patch
+"@terrazzo/plugin-tailwind": patch
+"@terrazzo/plugin-vanilla-extract": patch
+---
+
+Place `color-scheme` before CSS properties
--- a/packages/plugin-css/src/build/index.ts
+++ b/packages/plugin-css/src/build/index.ts
@@ -31,6 +31,12 @@ export default function buildFormat({
   if (rootTokens.length) {
     const rootRule: CSSRule = { selectors: [baseSelector], declarations: {} };
 
+    // add base color-scheme declaration first if configured
+    // (must be before other properties to ensure it appears first in output)
+    if (baseScheme) {
+      rootRule.declarations['color-scheme'] = { value: baseScheme };
+    }
+
     // note: `nestedQuery` was designed specifically for higher-gamut colors to
     // apply color-gamut media queries to existing selectors (i.e. keep the same
     // targets, and apply another nested layer of media query filtering based on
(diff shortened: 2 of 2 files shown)


drwpow (approved):
Great thank you!

drwpow:
Only test failure is from snapshots. If you could update those with `pnpm run test -u` that’d be great

packages/plugin-css/test/fixtures/type-border/want.css
```diff
@@ -45,15 +45,13 @@
 @media (prefers-color-scheme: light) {
   :root {
     --ds-border-red: 1px solid var(--ds-color-red);
-    --ds-border-thick: 2.5px solid var(--ds-color-border);
```
drwpow: Why did these get deleted?

packages/plugin-css/test/fixtures/utility-css/want.css
```diff
@@ -149,28 +149,24 @@
   text-transform: var(--ds-typography-base-text-transform);
 }
 
-.font-typography-size-lg {
-  font-size: var(--ds-typography-size-lg);
-}
-
-.font-typography-size-md {
-  font-size: var(--ds-typography-size-md);
+.font-typography-size-xs {
```
drwpow: Hm I think we have some snapshot failures still. I think the CSS classes shouldn’t have rearranged. Maybe try `pnpm run build` in your local branch and re-running the snapshots?

drwpow:
Update: you may want to change your approach somewhat. Before, we were sorting properties alphabetically. But with the addition of #589, we have to make sure that insert order is preserved. Really this is probably a signal we need to move away from an object, because objects should never try and preserve order, and for the most part they didn’t until #589.

Anyways, your approach can still work overall and I’m happy to merge this PR! But I think it’ll have to just not be a part of the `.sort()` method anymore since that’s removed.

packages/plugin-css/test/fixtures/type-border/want.css
nlemoine: 🤷‍♂️ Just ran the suggested command `pnpm run test -u`.

nlemoine:
@drwpow I rebased the PR and made the relevant changes to fit the new approach. 

There is one test failing on macos but I don't think it's related to this PR. 

Let me know if that looks good for you.

drwpow (approved):
Looks great, thank you! Agreed, the macos failure is unrelated. I’ll look into it separately. Thank you! 🙂

PR #591, opened 2025-11-20
Update legacy Terrazzo modes to a synthetic Resolver

drwpow:
## Changes

This lets folks use legacy Terrazzo modes with the new resolver syntax, without requiring any changes to the source tokens. This now always returns a `resolver` from `parse()`, which means both the old and new syntax will be supported (at least for all of 2.0).

There’s still some followup PRs and tests needed to fully support Resolvers, but the end is close! 🤞 

## How to Review

- Tests added; tests should pass.

### DIFF
--- a/packages/parser/test/resolver.test.ts
+++ b/packages/parser/test/resolver.test.ts
@@ -1,5 +1,5 @@
 import fs from 'node:fs/promises';
-import { parse as parseJSON } from '@humanwhocodes/momoa';
+import * as momoa from '@humanwhocodes/momoa';
 import stripAnsi from 'strip-ansi';
 import { describe, expect, it } from 'vitest';
 import defineConfig from '../src/config.js';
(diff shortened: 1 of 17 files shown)


packages/parser/test/resolver.test.ts
```diff
@@ -150,14 +150,92 @@ describe('Resolver module', () => {
...
+              type: 'set',
+              sources: [{ color: { blue: { '600': lightToken } } }],
+            },
+            {
+              type: 'modifier',
+              description: 'Automatically built from $extensions.mode',
+              contexts: {
+                light: [{ color: { blue: { '600': lightToken } } }],
+                dark: [{ color: { blue: { '600': darkToken } } }],
+              },
```
drwpow, line 210: 👆 Example of Resolver built automatically from `$extensions.mode`. Neat! 😎

www/src/pages/docs/2.0/integrations/js.md
```diff
@@ -67,8 +122,33 @@ export default defineConfig({
...
-| Name   | Type                | Description                                                                                                          |
-| :----- | :------------------ | :------------------------------------------------------------------------------------------------------------------- |
-| `js`   | `string \| boolean` | Set to a filename, or `false` to disable (default: `index.js`).                                                      |
-| `ts`   | `string \| boolean` | Set to a filename, or `false` to disable (default: `index.d.ts`) _Note: this can’t be enabled if `js` is disabled_). |
-| `json` | `string \| boolean` | Set to a filename, or `false` to disable (default: `false`).                                                         |
+| Name       | Type     | Description                               |
+| :--------- | :------- | :---------------------------------------- |
+| `filename` | `string` | Set to a filename (default: `tokens.js`). |
+
+## Migrating from 0.x
```
drwpow, line 129: This is more-or-less a roadmap for now. The JS plugin didn’t get updates in this PR, but “working backwards” and writing out the API I’m shooting for first, really helped clarify some upstream work in the parser to get all this ready.

It’ll be a little bit of a PITA to generate the JS for this, but the design is the hard part, and that’s shaping up nicely.

PR #592, opened 2025-11-21
feat: Resolver basic support

drwpow:
## Changes

PR to get basic Resolver support working. Still a lot of tests to add, but this should support the core usecases now.

TODO:

- `$extends` support
- Tests

## How to Review

- Tests updated

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@terrazzo/cli",
-  "version": "2.0.0-alpha.3",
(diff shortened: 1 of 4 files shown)


packages/parser/test/resolver.test.ts
```diff
@@ -245,7 +248,7 @@ describe('Resolver module', () => {
   });
 
   describe('Additional cases', () => {
-    it.skip('filesystem', async () => {
```
drwpow, line 248: First core test working! Now 999 to go…

Issue #593, opened 2025-11-22
Wonky bundling


mrginglymus:
There's some severe wonkiness with the bundling.

By default, rolldown will bundle everything. This means that deps declared under `dependencies` are being bundled, but still also being downloaded as dependencies. e.g. `@terrazzo/token-tools` is a prod dependency of `@terrazzo/plugin-css`, but is also inlined in the bundle.

Secondly, the the typescript stubs for `@terrazzo/plugin-css` include public types from `@terrazzo/parser`, which is _not_ included in `dependencies` and therefore relies on hoisting to be resolved correctly.

drwpow-figma:
Great catch! I’m a little embarrassed I didn’t notice this. Yes any/all improvements here are welcome (and we can also open issues on Rolldown too. Rollup did do a better job of prompting for externals)

mrginglymus:
>  Rollup did do a better job of prompting for externals

Yes rolldowns default behaviour of "Oh I couldn't find this it's probably fine" is...somewhat surprising!

PR #594, opened 2025-11-22
Example for #593

mrginglymus:
- re: #593

update config for `@terrazzo/plugin-css` to generate 'correct' output (no unnecessarily bundled js, no missing bundled types).

i'm surprised/disappointed arethetypeswrong didn't pick up on the transitive types

before:

```
<DIR>/index.d.ts      chunk │ size:   3.28 kB
<DIR>/index.js        chunk │ size: 145.43 kB
```

after:
```
<DIR>/index.d.ts      chunk │ size: 16.04 kB
<DIR>/index.js        chunk │ size: 16.35 kB
```

after, but also promoting `@terrazzo/parser` from dev to prod dep:

```
<DIR>/index.d.ts      chunk │ size:  3.28 kB
<DIR>/index.js        chunk │ size: 16.35 kB
```

### DIFF
--- a/packages/plugin-css/rolldown.config.ts
+++ b/packages/plugin-css/rolldown.config.ts
@@ -1,12 +1,15 @@
 import { defineConfig } from 'rolldown';
 import { dts } from 'rolldown-plugin-dts';
 
+import pkg from './package.json';
+
 export default defineConfig({
   input: {
     index: './src/index.ts',
   },
   platform: 'browser',
-  plugins: [dts()],
+  plugins: [dts({resolve: true})],
+  external: Object.keys(pkg.dependencies).flatMap(k => [k, new RegExp(`^${k}\/.*`)]),
   output: {
     dir: 'dist',
     format: 'es',


packages/plugin-css/rolldown.config.ts
```diff
@@ -1,12 +1,15 @@
 import { defineConfig } from 'rolldown';
 import { dts } from 'rolldown-plugin-dts';
 
+import pkg from './package.json';
```
drwpow-figma, line 4:
```suggestion
import pkg from './package.json' assert { type: 'json' };
```

Does this work? This wouldn’t affect anything but a) I wanna know if Rolldown processes this correctly and b) this could help newer versions of Node run this file without transpilation since Node can read TS now

mrginglymus: It works without, and it would need to be `with` not `assert` - but then you wouldn't need the `resolveJsonModule` flag in your tsconfig, although this file isn't actuall covered by any tsconfig files.

drwpow: Yeah it’s not a huge deal for now, just in a future world when everyone’s on newer Node versions that just understand TS, transpilation can be bypassed for less work and faster execution. If we ditched Rollup we could try this, but I see in your other PR it’s just using `node:fs` which works too

PR #595, opened 2025-11-22
Fix rolldown config

### LINKED ISSUE #593: Wonky bundling

mrginglymus:
## Changes

- fixes #593

## How to Review


mrginglymus:
A considerably simpler option is to drop tsdown, as it doesn't seem like it offers much benefit (on the current usage) over just compiling with tsc.

packages/cli/rolldown.config.ts
```diff
@@ -1,37 +1,8 @@
-import { defineConfig } from 'rolldown';
-import { dts } from 'rolldown-plugin-dts';
+import { defineConfig } from '../../rolldown';
```
drwpow, line 1: Suggestion: let’s rename this to `defaults()`, and revert back to Rolldown’s original defineConfig function. Then we can spread `... defaults()` and it’s easier to manipulate them, say, if we have to change one `external` or change a nested setting. The current implementation doesn’t allow modifying any defaults, only replacing it. Also at a skim it’s misleading if you didn’t realize `defineConfig()` is custom and secretly loading plugins and settings that aren’t apparent.

The duplication wasn’t all accidental; duplication in bundling can be much friendlier to edit because every package often can evolve to require its own bundling needs and settings. But in the duplication you pointed out the errors, which is a tradeoff. But if we make some tweaks we can improve consistency while allowing the bundling configs to still be individually-unique.

mrginglymus: I was hoping for a `mergeConfigs` functionality to make overrides easier...but alas nothing built in.

fwiw as a quick experiment I did try switching to using pure `tsc -b`, and it's significantly faster than rolldown - just a few seconds for a clean build, compared to >30s on main. If you're not actually bundling external packages, it's effectively the same output (as far as end users are concerned)

mrginglymus: But that is just my personal axe grinding about bundlers :)

drwpow: The reason I don’t like using tsc for popular packages is it doesn’t properly resolve extensions needed for native Node execution. You have to remember to write `.js` extensions on every file or your build is broken. It also can’t process any transformations needed for any non-JS files. We could probably get away with it for Terrazzo, but I just prefer leaning into Rolldown and managing less. I’ve just made all the mistakes one can shipping popular packages with only tsc and have hit all its limits 🙃

mrginglymus: With `moduleResolution: NodeNext` and solid type checking you should catch all of those extensions issues. I've got a [branch here](https://github.com/terrazzoapp/terrazzo/compare/main...mrginglymus:terrazzo:tsc) - if you remove one of the extensions, it will fail the build.

Rolldown plugin dts has some serious shortcomings though. It can't handle `@types/` packages, `typesVersions` references, and generates phantom exports - all of which it does silently. Which is better than roll_up_ plugin dts, which would sometimes throw some executable javascript code in your `d.ts` files for fun.

packages/json-schema-tools/package.json
```diff
@@ -42,11 +42,9 @@
   "peerDependencies": {
     "@humanwhocodes/momoa": "^3.0.0"
   },
-  "dependencies": {
-    "@humanwhocodes/momoa": "catalog:"
```
mrginglymus, line 47: This was lited in both peer and prod deps, which was presumably in error.

rolldown.ts
```diff
@@ -10,7 +10,11 @@ export const defineConfig = (options: RolldownOptions): RolldownOptions => {
     {
       platform: 'browser',
       plugins: [dts({ resolve: true })],
-      external: Object.keys(pkg.dependencies ?? {}).flatMap((k) => [k, new RegExp(`^${k}/`)]),
+      external: [
+        'yaml-to-momoa',
```
mrginglymus: I feel like this ought to be an optional peer dependency in the places it's currentl referenced...but for now, just marked it as always external

mrginglymus:
green \o/

packages/cli/rolldown.config.ts
drwpow: Ugh you have some good points. Given all that I’d be OK with simply ripping out Rolldown in favor of TSC + NodeNext. We still have to use Roll(up|down) for frontend libraries like `@terrazzo/tiles` or any library that ships non-JS/TS like `.css` files (building other filetypes require Rolldown anyway, so you might as well use one pipeline for everything). But as far as this project goes, all the high-traffic packages 
are pure JS/TS and can be handled by TSC.

I haven’t been following typescript-go closely but I’d probably just wait till that was stable to kick the tires. There are many problems that pique my interest that I love working on, but generating TS -> JS/DTS is not one of them 😅

rolldown.ts
drwpow: Oh interesting! I’ve never tried optional peer dependencies but that seems correct.

mrginglymus: It's a weird one because it _will_ install it (unless you specifiy no optional in your package manager)...it'll just be ok if it fails to install for some reason. Moving it to peerDeps will mean it'll get _installed_ regardless, I think....at which point, why not just have it a dep?

[Python](https://peps.python.org/pep-0508/#extras) and [Rust](https://doc.rust-lang.org/cargo/reference/features.html) ahead of the game, again...

packages/cli/rolldown.config.ts
mrginglymus: Still waiting on it being considered acceptable to just ship typescript files....then we don't have to worry about skipLibCheck or any of that nonsense.

I personally like just using tsc when I can, but I also recognise the benefits of rolldown. I'm happy to contribute a PR that converts this repo to tsc (where appropriate), or get this one polished - your call :)

rolldown.ts
drwpow: For more context, the reason `yaml-to-momoa` is a separate, manual install for `@terrazzo/parser` is for those few users who are building client-side tools, `yaml-to-momoa` introduces the `yaml` dep which is extremely heavy. For Node CLIs, though, it’s trivial, and we install it automatically with `@terrazzo/cli`.

To your point, though, we could just install it automatically and that would be fine, since they still have to manually load `yaml-to-momoa` in the parser API. So I wouldn’t mind it being a dep or peerDep that gets automatically installed.

packages/cli/rolldown.config.ts
drwpow: Yeah, sorry I wasn’t being clear. I’m open to either path, whatever you feel is less work. This seems like it’s in a good state as-is, and I’d be happy to merge it. If you’d like we could merge this, since it just improves the bundling without any downsides, and then just convert packages to `tsc` one-by-one after this merges. Sound good?

mrginglymus: Sounds like a plan

rolldown.ts
mrginglymus: Cool, I'll convert it to a peer dep and see what happens (:

mrginglymus:
@drwpow is this mac os test a known flaky or something I've broken?

drwpow:
> @drwpow is this mac os test a known flaky or something I've broken?

It’s known flaky. It has something to do with updating Vitest 3 → 4. There’s a job that tests `build --watch` which is a little complex, but up till now had been working no problem, and now there’s some hanging process that doesn’t shut down properly but only in the macos runner. I’ve poked at it a few times this week but haven’t solved it; I may just disable the watch test on mac 🤷.

(no nothing you broke 😛)

PR #597, opened 2025-11-23
Add missing format commands

### LINKED ISSUE #596: `lint` and `format` disagree
After changing a few files, `lint` said some files needed formatting. So I ran `format`, and it didn't change the files `lint` said needed changing but it did change some others.

You can repro this on `main` by running `format` right now.

mrginglymus:
- fixes #596

A couple of packages were missing format commands.

### DIFF
--- a/www/src/components/docs-footer-nav.astro
+++ b/www/src/components/docs-footer-nav.astro
@@ -5,13 +5,13 @@ const nav = (Astro.url.pathname.includes('/docs/2.0') ? DOCS_NAV_V2 : DOCS_NAV).
   section.url ? section : (section.items ?? []),
 );
 
-let _prev: { url?: string; title?: string } | undefined;
-let _next: { url?: string; title?: string } | undefined;
+let prev: { url?: string; title?: string } | undefined;
+let next: { url?: string; title?: string } | undefined;
 
 for (let i = 0; i < nav.length; i++) {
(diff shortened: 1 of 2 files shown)


www/src/components/docs-footer-nav.astro
```diff
@@ -5,13 +5,13 @@ const nav = (Astro.url.pathname.includes('/docs/2.0') ? DOCS_NAV_V2 : DOCS_NAV).
   section.url ? section : (section.items ?? []),
 );
 
-let prev: { url?: string; title?: string } | undefined;
-let next: { url?: string; title?: string } | undefined;
+let _prev: { url?: string; title?: string } | undefined;
```
drwpow: We need to revert this. For some reason BiomeJS has errors parsing Astro files and thinks vars are unused when they aren’t (they’re just used in the template). It’s 50/50 though because Biome does catch some things! So it’s felt better than nothing, but not reliable enough to really enforce.

Should we just turn off linting/formatting for `.astro` files altogether maybe? Or just find the overrides in Biome to be less aggressive here? I’ve just procrastinated on the decision honestly

mrginglymus: Huh - I assumed it was a windows bug but turns out biome do recommend turning this off for astro files:

https://biomejs.dev/internals/language-support/#linting-html-ish-languages

I'll try adding that config to this PR (and reverting this file change)

drwpow (approved):
Thank you!

PR #615, opened 2025-12-14
docs: Fix code issue in plugin example

shivajivarma:


drwpow-figma (approved):
Good catch, thank you!

PR #620, opened 2025-12-16
Fix return type of findNode, add logger.assert

mrginglymus:
## Changes

`findNode` was casting to not-undefined, which was not safe. It was also unsafely casting to a specific momoa Node based on the call site, which is unsafe.

Removed the unsafe casting (or moved it to the call site in one case).

Added a `logger.assert` method that asserts a value is truthy. This means we can remove many of the non-null assertions that follow previous calls to `logger.error`.

Note that there are some headaches around assertion functions that require the logger class to be explicitly typed: https://github.com/microsoft/TypeScript/issues/47945

edit: added another assertion for subtyping

### DIFF
--- a/packages/parser/test/lint.test.ts
+++ b/packages/parser/test/lint.test.ts
@@ -1,6 +1,6 @@
 import { describe, expect, test } from 'vitest';
 
-import { defineConfig, type Logger, parse } from '../src/index.js';
+import { defineConfig, type LogEntry, Logger, parse } from '../src/index.js';
 import {
   A11Y_MIN_CONTRAST,
   A11Y_MIN_FONT_SIZE,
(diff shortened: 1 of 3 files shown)


packages/parser/src/logger.ts
```diff
@@ -110,6 +110,13 @@ export default class Logger {
     }
   }
 
+  /** Asserts a given value passes the provided test; if it isn't, log errors (and exit) */
+  assertTest<I, T extends I>(value: I, test: (value: I) => value is T, ...entries: LogEntry[]): asserts value is T {
```
drwpow, line 114: This is a neat utility but I don’t want to add this as part of a logger API everywhere, especially since we expose this to plugins and outside the package. I’m mostly looking to mirror Rollup’s minimal logging utility, not deliver something full-featured (that we have to support as an API).

packages/parser/src/parse/process.ts
```diff
@@ -101,8 +101,6 @@ export function processTokens(
   function flatten$extends(node: momoa.ObjectNode, chain: string[]) {
     const memberKeys = node.members.map((m) => m.name.type === 'String' && m.name.value).filter(Boolean) as string[];
 
-    let extended: momoa.ObjectNode | undefined;
```
drwpow, line 104: Oh good catch 🙏. I thought Biome would have caught this; this was from an earlier refactor

drwpow (commented):
Thanks for the fix! I’m overall in favor of this change but I don’t want to extend the global logger API at this time. Let’s just throw an error inline while keeping the overall change.

mrginglymus:
Gotcha, completely understandable. We could have an internal utilty to do this:

```ts
export function assert(condition: unknown, logger: Logger, ...mesages: LogMessage[]): asserts condition
```

to avoid polluting the logger?

drwpow-figma:
> Gotcha, completely understandable. We could have an internal utilty to do this:
> 
> ```ts
> export function assert(condition: unknown, logger: Logger, ...mesages: LogMessage[]): asserts condition
> ```
> 
> to avoid polluting the logger?

Yup! I’d be fine with that.

mrginglymus:
Presumably we can't use `node:assert` because browser support? Was thinking of a top level util to hoover up assertion errors and plumb them into the loggerr...but the util is probably best

mrginglymus:
Did this in #622 as it was easier to start from scratch

drwpow:
Superceded by #622

PR #623, opened 2025-12-17
Token transform

mrginglymus:
## Changes

Reworking of #621 

I realised that the default `transformAlias` does throw errors on undefined:

https://github.com/terrazzoapp/terrazzo/blob/8cb471e3cd8468fbf8d47c454344c00fb0a78d83/packages/token-tools/src/css/lib.ts#L6-L9

However, when a user-provided function is given, this assertion is lost.

Instead of having to update every transform, update the entrypoint to wrap any user-provided `transformAlias` with an assertion. This means we can burn a lot more `!`.

## How to Review

### DIFF
--- a/packages/token-tools/src/css/lib.ts
+++ b/packages/token-tools/src/css/lib.ts
@@ -1,9 +1,14 @@
 import type { TokenNormalized } from '../types.js';
 
+/**
+ * Function that generates a var(…) statement
+ * Will throw error if token not provided
+ * */
+export type StrictIDGenerator<T = TokenNormalized> = (token: T | undefined) => string;
 /** Function that generates a var(…) statement */
 export type IDGenerator<T = TokenNormalized> = (token: T) => string;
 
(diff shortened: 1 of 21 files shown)


packages/token-tools/src/css/lib.ts
```diff
@@ -1,9 +1,14 @@
...
+ * Will throw error if token not provided
+ * */
+export type StrictIDGenerator<T = TokenNormalized> = (token: T | undefined) => string;
 /** Function that generates a var(…) statement */
 export type IDGenerator<T = TokenNormalized> = (token: T) => string;
 
-export function defaultAliasTransform(token: TokenNormalized) {
+export function defaultAliasTransform(token: TokenNormalized | undefined) {
   if (!token) {
     throw new Error('Undefined token');
```
mrginglymus, line 13: This still isn't a huge improvement as we don't know what the token originally was...

PR #624, opened 2025-12-17
Make tokenset throw error when accessing invalid token

mrginglymus:
Another approach to #623 - wrap `tokenSet` in a proxy that logs an error using the provided logger when attempting to access a non-existant token.

### DIFF
--- a/packages/parser/src/parse/index.ts
+++ b/packages/parser/src/parse/index.ts
@@ -100,7 +100,18 @@ export default async function parse(
   }
 
   return {
-    tokens,
+    tokens: new Proxy(tokens, {
+      get(target, p: string): any {
+        const resolved = target[p];
+        if (!resolved) {
+          logger.error({
+            group: 'parser',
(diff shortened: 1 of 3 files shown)


mrginglymus:
What I think would be even better is instead of passing in the tokenSet, pass in an alias resolver function from the parser:

```ts
  resolveAlias(token: TokenNormalized): TokenNormalized | undefined {
    const alias = token.aliasChain?.[0];
    if (!alias) return;
    
    const resolved = tokensSet[alias];
    if (!resolved) {
      this.logger.error({ 
        group: 'parser',
        label: 'resolver',
        message: 'Could not resolve alias `${alias}`,
        node: token.source.node
      });
    }
    return resolved;
  }
```

but that would be a more substantial change to...well, everything (:

mrginglymus:
e.g. something like this: https://github.com/mrginglymus/terrazzo/pull/new/resolve-alias

drwpow:
Ended up merging a fix for this in #664!

Issue #629, opened 2025-12-27
Transform plugin-tailwind CSS variable names to kebab-case


9rotama:
### Versions

- @terrazzo/cli 0.10.5
- @terrazzo/plugin-css 0.10.4
- @terrazzo/plugin-tailwind 0.3.3

### Node.js version

22.17.0

### OS + version

macOS 29.1

### Description

When using Tailwind plugin, I want to convert token names to kebab-case uniformly, similar to CSS Plugin. Currently, we are experiencing the following issue with the token name format.

- Uppercase letters being included (Using uppercase for CSS variables and Tailwind theme variables is a rare case for me personally.)
- Token names that include spaces or other characters invalid in CSS cannot be used as variables.

### Reproduction

`terrazzo.config.mjs`:
```js
export default defineConfig({
  tokens: ['./tokens/test.tokens.json'],
  plugins: [
    css({
      filename: 'css-tokens.css'
    }),
    tailwind({
      filename: 'tailwind-tokens.css',
      theme: {
        color: ['Color.*']
      }
    })
  ],
  ...
});
```

`test.tokens.json`:
```json
{
  "Color": {
    "Background Blue": {
      "50": {
        "$type": "color",
        ...
      }
    }
  }
}
```

Execute `npx tz build`, then the following will be output:

```css
@import "tailwindcss";

@theme {
  --color-Background Blue-50: color(srgb 0.9098039215686274 0.9450980392156862 0.996078431372549);
}
```

### Expected result

CSS plugin outputs the following code:

```css
:root {
  --color-background-blue-50: color(srgb 0.9098039215686274 0.9450980392156862 0.996078431372549);
}
```
In line with that, Tailwind plugin is expected to output code like this:
```css
@import "tailwindcss";

@theme {
  --color-background-blue-50: color(srgb 0.9098039215686274 0.9450980392156862 0.996078431372549);
}
```

### Extra

- [x] I’m willing to open a PR (see [CONTRIBUTING.md](https://github.com/terrazzoapp/terrazzo/blob/main/CONTRIBUTING.md))

PR #630, opened 2025-12-30
Transform plugin-tailwind CSS variable names to kebab-case

9rotama:
## Changes

In the fix for issue #629, we will change the variable names output by the Tailwind plugin to kebab-case.
- Using the makeCSSVars function from @terrazzo/token-tools, we will convert the localID created during transformation to kebab-case.
- Affected fixtures will be corrected.

## How to Review

Please let me know if it needs to be opt-in, as it is likely a breaking change.


packages/plugin-tailwind/test/fixtures/primer/want.css
```diff
@@ -61,37 +61,37 @@
   --color-red-7: #82071e;
   --color-red-8: #660018;
   --color-red-9: #4c0014;
-  --color-yellow-0-: #fff8c5;
```
9rotama, line 64: When a token name is specified in the config with the theme property instead of a wildcard, the hyphen will no longer be added.

drwpow:
This fix looks great! If you could add a changeset (see comment) I’d be happy to approve & merge.

PR #632, opened 2026-01-06
Update CSS transform return type to include undefined or null

aduth:
## Changes

Updates the TypeScript type for the CSS plugin's `transform` option to specify `undefined` or `null` as valid return types.

This is consistent with how the return value is used in the code itself, which explicitly checks for `undefined` or `null` values before applying the transformed value:

https://github.com/terrazzoapp/terrazzo/blob/a15d358497ed63289bcde3d296a0fd083014b133/packages/plugin-css/src/index.ts#L50-L51

Similarly, [documented examples for the `transform` function](https://terrazzo.app/docs/integrations/css/#transform) include code snippets which don't have return values in every path, which currently results in TypeScript errors in projects using TypeScript type-checking:

```ts
export default defineConfig({
  plugins: [
    css({
      transform(token, mode) {
        if (token.id === "token.i.want" && mode === ".") {
          return "my-custom-value"; // generates `--token-i-want: my-custom-value;`
        }
      },
    }),
  ],
});
```

```
Type '(token: TokenNormalized) => "my-custom-value" | undefined' is not assignable to type '(token: TokenNormalized, mode: string) => string | Record<string, string>'.
  Type 'string | undefined' is not assignable to type 'string | Record<string, string>'.
    Type 'undefined' is not assignable to type 'string | Record<string, string>'.ts(2322)
```

## How to Review

TypeScript type-checking should pass. Following contributing guidance, I'd expect this should be validated through `pnpm lint` .

### DIFF
--- a/.changeset/fix-transform-return-type.md
+++ b/.changeset/fix-transform-return-type.md
@@ -1,0 +1,5 @@
+---
+"@terrazzo/plugin-css": patch
+---
+
+Fix `transform` option TypeScript type to allow `undefined` or `null` return values


ciampo:
cc @drwpow

drwpow:
This change is fine. Could you add a changeset as indicated by the comment? And I can approve & merge!

aduth:
@drwpow Sure thing! Added in 65a1897e04e4c596b98c2e1f8adf89f317ae52e9.

PR #642, opened 2026-01-22
Per permutation include/exclude token filter

mrginglymus:
POC of flat resolver output

### DIFF
--- a/www/src/pages/docs/2.0/integrations/css.md
+++ b/www/src/pages/docs/2.0/integrations/css.md
@@ -257,6 +257,60 @@ You control the wrapper CSS, so check for mistakes! If using `@media` queries, r
 
 If you inspect the output CSS, you may find more variables than expected in the media queries. This is necessary the way CSS works: if a CSS variable is an alias of another, when the base value changes, all aliases must be redeclared otherwise they are referencing the old value in the parent scope. At first glance, this seems like a bug, with variables being redeclared with the same values, but in actuality it’s necessary so your mode selectors cascade correctly.
 
+### Selective token output
+
+The CSS plugin can filter which tokens are output to CSS. This is useful if you want to generate CSS for only a subset of your tokens, and
+can be applied at the plugin or permutation level using `include` and `exclude` token globs.
+
+#### Selective output at the plugin level
+
+:::code-group
+
+```ts [terrazzo.config.ts]
+import { defineConfig } from "@terrazzo/cli";
+import css from "@terrazzo/plugin-css";
+
+export default defineConfig({
+  plugins: [
+    css({
+      include: ['primitives.*'], // include only primitives
+      exclude: ['primitives.typography.*'], // except typography primitives
+    }),
+  ],
+});
+```
+:::
+
+#### Selective output at the permutation level
+
+:::code-group
+
+```ts [terrazzo.config.ts]
+import { defineConfig } from "@terrazzo/cli";
+import css from "@terrazzo/plugin-css";
+
+export default defineConfig({
+  plugins: [
+    css({
+      permutations: [
+        {
+          input: {},
+          prepare: (css) => `:root {\n  ${css}\n}`,
+          include: ['primitives.*'], // include only primitives in this permutation
+        },
+        {
+          input: { mode: 'light' },
+          prepare: (css) => `.light {\n  ${css}\n}`,
+          exclude: ['primitives.*'], // include everything but primitives in this permutation
+        },
+      ],
+    }),
+  ],
+});
+```
+
+:::
+
 ### Utility CSS
 
 Using the [Tailwind integration](/docs/integrations/tailwind) isn’t necessary if you want to just have utility classes generated from your tokens. You can generate Tailwind-like utility CSS with minimal config.


packages/plugin-css/test/fixtures/flat-contexts/terrazzo.config.ts
```diff
@@ -0,0 +1,67 @@
...
+  filename: resolverPath,
+  src: await fs.readFile(resolverPath, "utf-8"),
+});
+
+const allTokens = new Set(Object.keys(tokens));
+/**
+ * All token names defined in resolver sets
+ */
+const rootTokens = Object.values(resolver.source.sets || {}).reduce((acc, set) => {
+    return acc.union(set.sources.reduce((acc, source) => {
```
mrginglymus: Having an API for this on the resolver could prove super useful. Or is there a utility somewhere that can recursively extract keys from a resolved source?

packages/plugin-css/src/build.ts
```diff
@@ -104,6 +110,9 @@ export default function buildCSS({
         // redeclare aliases so they have the correct scope
```
mrginglymus, line 118: It's this block where I feel terrazzo might be overstepping its responsibilities.

Redeclaring aliasses from modifiers may result in output that prevents using the classes as indepenent toggles for that modifier. Whilst it guarantees that the alias vars will resolve to _something_, it also significantly increases the chances that they will resolve to the _wrong_ thing.

However, respecting the include/ignore options here will fix that problem, and the possibility of unresolved aliases owing to user error is already signposted in the docstrings for include/exclude.

drwpow: Agree. It made more sense in `$extensions.mode` which have a 2-tiered system, but for permutations we have to redeclare anything because people may be formulating 3+ tier systems, or they may be mutually-exclusive. We should just print out everything for every permutation so generation works for everyone.

packages/plugin-css/src/build.ts
```diff
@@ -104,6 +110,9 @@ export default function buildCSS({
...
         for (const alias of aliasTokens) {
           if (alias.localID && typeof alias.value === 'string') {
+            if (!include(alias.id) || ignore?.(alias.id) || exclude?.(alias.id)) {
```
drwpow: `ignore` was not an intentional departure from include/exclude terms. I’d probably prefer to just include/exclude everywhere

packages/plugin-css/src/build.ts
```diff
@@ -58,7 +59,12 @@ export default function buildCSS({
         rec2020: [] as CSSDeclaration[],
       };
 
+      const ignore = p.ignore ? wcmatch(p.ignore) : undefined;
+      const include = p.include ? wcmatch(p.include) : () => true;
```
drwpow: (Note unrelated to your PR) these `wcmatch` setups are fast but do have some cost to them, and we see perf tank if we redeclare these per-token just due to the sheer number of tokens. Keeping them as high up as possible like this is ideal, and not re-calling them if the options didn’t change

packages/plugin-css/test/fixtures/flat-contexts/terrazzo.config.ts
drwpow: The tricky part here is some folks may declare sets inline in `resolutionOrder` (same for modifiers). We do NOT hoist up inline sets/modifiers into those top-level keys because that could break `$refs`. But we DO normalize everything in `resolutionOrder` because that is safer. So if you want everything, you should iterate over that instead.

I wouldn’t mind adding `resolver.sets` and `resolver.modifiers` to the top-level that’s perhaps a little more flattened/normalized than `resolver.source`. The tricky part is finding a balance between keeping metadata and implementation details in place as the user wrote them, without making them parse it (like some folks may be trying to parse their own `$ref`s—is that good or bad? Is that a core usecase to support?)

packages/plugin-css/test/fixtures/flat-contexts/terrazzo.config.ts
```diff
@@ -0,0 +1,65 @@
...
+}`
+            },
+            include: [...modifierTokens]
+        });
+    })
+})
+
+export default defineConfig(
+    {
+        plugins: [css({filename: 'actual.css', permutations})],
```
drwpow, line 47: Suggestion: for a test fixture, this test should hardcode the `input` because it’s impossible to grok what this list is/what’s expected.

However there could be value in testing some programmatic aspects of dealing with resolvers, but that would likely live in parser tests (we don’t need to test the programmatic creation of an object in plugin-css)

mrginglymus: That's a fair point - I think I wanted to double check the type change made in the first commit.

That being said, the duplication of the `prepare` hook would be a bit sad. I can certainly rewrite it in a slightly clearer way, let me see what that looks like...

drwpow: Yeah it’s the fundamental difference between test fixtures vs real code—it’s really more about being able to see input/output snapshots here without using too many braincells (I don’t have many :P)

packages/plugin-css/test/fixtures/flat-contexts/terrazzo.config.ts
```diff
@@ -0,0 +1,85 @@
...
+const resolverPath = new URL("./resolver.json", import.meta.url);
+
+const {resolver} = await parse({
+    filename: resolverPath,
+    src: await fs.readFile(resolverPath, "utf-8"),
+});
+
+
+/**
+ * These three would be very useful to have in the resolver API
```
mrginglymus: *five :P

packages/plugin-css/test/fixtures/flat-contexts/terrazzo.config.ts
```diff
@@ -0,0 +1,85 @@
...
+    }, new Set<string>())]
+}
+
+function extractTokensFromContexts(contexts: ResolverModifierNormalized['contexts']): string[] {
+    return Object.values(contexts).flatMap(groups => extractTokenNames(...groups))
+}
+
+function iterateModifiers(modifiers: ResolverSourceNormalized['modifiers']): (readonly [ResolverModifierNormalized, string])[] {
+    return Object.values(modifiers || {}).flatMap(modifier => Object.keys(modifier.contexts).map(context => [modifier, context] as const))
+}
```
mrginglymus: @drwpow these are the utilities that I will need to use to achieve the output I want. They seem like they might be useful for other users too (to varying degrees).

I will switch this fixture to static data at some point, just wanted to highlight the utility functions.

packages/plugin-css/src/build.ts
mrginglymus: I was just browsing the docs for `wildcard-match` - should we not have '.' as the separator?

```ts
    // Current behaviour 
    expect(wcmatch('foo.*.bar')('foo.yay.hoopla.bar')).toBeTruthy();
    expect(wcmatch('foo.*.bar')('foo.yay.bar')).toBeTruthy();
    // Possibly more expected behaviour
    expect(wcmatch('foo.*.bar', '.')('foo.yay.hoopla.bar')).toBeFalsy();
    expect(wcmatch('foo.*.bar', '.')('foo.yay.bar')).toBeTruthy();
```

mrginglymus: see https://github.com/terrazzoapp/terrazzo/pull/643

drwpow: Oh that’s a good call! Yes let’s start using `.` as the separator

packages/plugin-css/test/fixtures/flat-contexts/terrazzo.config.ts
drwpow: Yeah. I’m not against adding these, but I still think I’m lacking some context in when to use them, and where. Writing up a docs stub about some common usecases would help a lot (even if it’s a GitHub Gist and not a PR with docs, though the latter would be great!)

mrginglymus:
@drwpow updated to:

1. change permutations to include/exclude (instead of include/ignore)
2. add 'include' to root options (for consistency)
3. simplified test to remove sniffing logic

mrginglymus:
@drwpow done!

PR #643, opened 2026-01-23
Extract token matcher to utility in token-utils

mrginglymus:
## Changes

This ensures that `color.*.60` does not match `color.blue.tint.60`.

This is a potentially breaking change, but I think it matches more accurately with expectations of what a glob matcher will do on a `.` separated string.

## How to Review


packages/plugin-sass/src/build.ts
```diff
@@ -58,8 +58,10 @@ export default async function build({ getTransforms, options }: BuildParams): Pr
 
   const tokenValuesMap = root.createMap();
 
+  const exclude = options?.exclude ? getTokenMatcher(options?.exclude) : undefined;
+
   for (const token of getTransforms({ format: FORMAT_ID, id: '*', mode: '.' })) {
-    if (getTokenMatch(token.token.id, options?.exclude ?? [])) {
+    if (exclude?.(token.token.id)) {
```
mrginglymus, line 64: As noted in the other PR, this may have a small positive impact on perf

packages/token-tools/src/id.ts
```diff
@@ -7,13 +7,9 @@ export function isAlias(value: string): boolean {
...
-/** Same as isTokenMatch but returns the matching pattern */
-export function getTokenMatch(tokenId: string, globPatterns: string[]): string | undefined {
-  for (const pattern of globPatterns) {
-    if (wcmatch(pattern)(tokenId)) {
-      return pattern;
-    }
-  }
+/** Create a token matcher function using . separated glob patterns */
+export function getTokenMatcher(globPatterns: string | string[]): (tokenId: string) => boolean {
+  return wcmatch(globPatterns, '.');
```
mrginglymus, line 12: Presumably a leftover from previous usage, the return value of this function was only used in truthiness testing so it could be simplified to this.

mrginglymus:
@drwpow this sits on top of #642 to pre-empt the conflicts.

I've updated all existing globs that ended `.*` to `.**` to preserve existing behaviour.

PR #644, opened 2026-01-23
Type tweaks

mrginglymus:
## Changes

Fix a couple of what _appear_ to be mistyped types

## How to Review


drwpow (approved):

drwpow:
I trust you here—I’m not sure! But we do have a lint error that needs to be fixed

mrginglymus:
I'm a bit less confident in the other one so I've backed it out for now. Will hunt down those other erorrs.

edit: less confident in that one too...some thinkings required....

Issue #543, opened 2025-09-12
Add option in plugin-tailwind to control `@import "tailwindcss";`


Hashversion:
### Summary

Currently, @terrazzo/plugin-tailwind outputs a theme file that always includes `@import "tailwindcss";` at the top. In some setups (monorepos, shared tokens packages, or when consumers already import Tailwind globally), this creates duplication or unwanted imports.

It would be useful to make this behavior configurable in the plugin, so projects can decide whether or not to include the `@import "tailwindcss";` line.

### Proposal & prior art

1. Syntax

Add a new plugin option in terrazzo.config.js when configuring the Tailwind plugin. For example:

```js
tailwind({
  filename: "tailwind-theme.css",
  includeTailwindImport: true // default
})
```

true → output file includes `@import "tailwindcss";` (current behavior)
false → omit the import (useful if Tailwind is already imported elsewhere)

2. Alternatives

Option A (recommended): Boolean flag (includeTailwindImport), simple, minimal change.
Option B: Always omit the import, breaking change for existing users who expect it.

3. Breaking change?

If we default to true, this is not breaking (preserves current behavior).
Only users who want to suppress the import would opt into the new flag.

4. Prior art

Tools like Style Dictionary allow flexible control of output templates.
PostCSS/Tailwind setups often separate concerns: apps import Tailwind globally, while design tokens packages only provide variables.

5. Expected output

Current output (always):

```css
@import "tailwindcss";

@theme {
  --color-primary: #3b82f6;
}
```

With `includeTailwindImport: false:`

```css
@theme {
  --color-primary: #3b82f6;
}
```

### Extra

- [ ] I’m willing to open a PR (see [CONTRIBUTING.md](https://github.com/terrazzoapp/terrazzo/blob/main/CONTRIBUTING.md))

davejsdev:
The use case makes sense! 

Thinking about this more holistically, we might consider a more generalized name for this config value to cover potential cases where other imports may need to be included/excluded from a config.

This is just the first thought I had, feel free to disagree.

```ts
tailwind({
  filename: "tailwind-theme.css",
  includeImports: true // default
})
```

or maybe something more configurable (perhaps overkill):

```ts
tailwind({
  filename: "tailwind-theme.css",
  imports: {
    "tailwindcss": true,  // default
    "tw-animate-css": true, // additional import
  }
})
```

drwpow:
> or maybe something more configurable (perhaps overkill):

[...]

PR #646, opened 2026-01-25
feat: Add an option in plugin-tailwind to set template CSS

9rotama:
## Changes

For issue #543, I will make it possible to specify custom template CSS as an option.

- Accept a path to the template CSS and inject the theme at the `@terrazzo-slot` position within the file.
- Add header comments to the output CSS.
- Add test fixtures for the template options.

## How to Review

Even when no template is specified (default behavior), a comment will now be included at the top of the output CSS.

### DIFF
--- a/packages/plugin-tailwind/src/lib.ts
+++ b/packages/plugin-tailwind/src/lib.ts
@@ -59,5 +59,6 @@ export function applyTemplate(template: string, generatedTheme: string): string 
   if (!template.includes(TERRAZZO_SLOT)) {
     throw new Error(`Template must contain "${TERRAZZO_SLOT}" directive`);
   }
-  return template.replace(TERRAZZO_SLOT, generatedTheme).replace(/\n{3,}/g, '\n\n');
+  // Replace slot and any following newlines (CRLF or LF) with theme + single newline
+  return template.replace(/@terrazzo-slot;[\r\n]+/, `${generatedTheme.trimEnd()}\n\n`);
 }
--- a/packages/plugin-tailwind/test/fixtures/template/template.tz.css
+++ b/packages/plugin-tailwind/test/fixtures/template/template.tz.css
@@ -2,6 +2,10 @@
 
 @terrazzo-slot;
 
+
+
+
+
 /* Custom styles */
 .custom-class {
   color: red;


packages/plugin-tailwind/src/index.ts
```diff
@@ -62,14 +68,27 @@ export default function pluginTailwind(options: TailwindPluginOptions): Plugin {
...
-        output.push('}', '');
+        themeOutput.push('}', '');
+      }
+      const generatedTheme = themeOutput.join('\n');
+
+      // Build final output
+      let finalOutput: string;
+      if (options.template) {
+        const templateUrl = new URL(options.template, cwd);
+        const templateContent = fs.readFileSync(fileURLToPath(templateUrl), 'utf8').replace(/\r\n/g, '\n');
```
9rotama: Normalize line endings in the read CSS file to LF so the test-windows workflow passes.

drwpow: This is totally OK to do! But **let’s do this in the test, not the runtime code**.

This plugin should preserve whatever line endings the user wants. Also, this could slow the plugin down unnecessarily, especially if they do want the `\r\n` endings in the end and have to transform them back (some people could also have gigantic templates, and this would be a noticeable slowdown)

packages/plugin-tailwind/test/fixtures/template/want.css
```diff
@@ -0,0 +1,17 @@
...
+ * ------------------------------------------- */
+
+@import "tailwindcss";
+
+@theme {
+  --color-primary: #3366cc;
+  --color-secondary: #cc3366;
+}
+
+
```
drwpow: Could we get rid of this extra line break here? This seems minor but folks will ask about it 😓 (they do seem to scrutinize the CSS output quite frequently)

drwpow (approved):
Amazing! Thanks so much! Sorry for nit-picking the CSS output, it’s just a common complaint, sadly 🙂

drwpow:
Sorry in advance for being a little cryptic, but just a heads up that as of #666 and the latest beta version, I’m realizing we’re hitting a wall trying to generate a full Tailwind config just from `terrazzo.config.ts`. So I’d like to go all-in on templates and **require** folks always use a template file 🙂.

I’ll put up a PR for feedback, as it will involve some _slight_ tweaks but overall very much inline with the direction you set. Just building off this and allowing what is hopefully more control and less config. I may go ahead and release some changes in another `beta` version just to get testing & feedback but I’d still love feedback all-the-same before we release the final version.

Will tag you on the upcoming PR!

drwpow:
Here’s the overall direction I’d love feedback on: https://github.com/terrazzoapp/terrazzo/pull/670.

As stated, I think I am going to release this **only so folks can test it,** but I’d love feedback if/when you’re able to provide it. Please preview `@terrazzo/plugin-tailwind@alpha`, see if it works as-expected, and either comment on the PR or open up a new issue. Thanks! 🙂

PR #647, opened 2026-01-27
Add per-permutation custom transform with full options

mrginglymus:
## Changes

This enables users to have full access to the context required for custom transforms in permutations.

This could be added to the root `transform` option, though it would clash with the existing 'mode' arg before that's removed.


mrginglymus:
Hmm, this doesn't work on composite tokens though - we really need a way to hook `transformAlias`...

mrginglymus:
Aha:

```ts
const transform: Permutation["transform"] = (token, options) => {
  const transformAlias = (token: TokenNormalized): string => {
    const defaultAlias = (options.transformAlias ?? defaultAliasTransform)(
      token,
    );
    if (shouldBypassAlias(token)) {
      const resolved = options.tokensSet[token.id];
      // just to avoid a non-null assertion
      if (!resolved) return defaultAlias;
      const transformed = transformCSSValue(resolved, {
        ...options,
        transformAlias,
      });
      // Users would probably want to make this a noisy error
      return typeof transformed === "string" ? transformed : defaultAlias;
    }
    return defaultAlias;
  };

  return transformCSSValue(token, { ...options, transformAlias });
};
```

This PR enables the above to be used in userland

packages/plugin-css/src/lib.ts
mrginglymus:
```ts
    modeOrOptions: string | TransformCSSValueOptions,
```

Would be the ideal change here...

drwpow: Yeah at this point I’m OK having this be a breaking change, and not have mode.

We have sort of a “connector” to resolvers where we create a virtual modifier named `tzMode`. So it would be something like:

```diff
- transform(token, mode) {
-   if (mode === "dark") {
+ transform(token, { permutation }) {
+   if (permutation.tzMode === "dark") {
```

We can make a breaking API change so long as the migration path is straightforward like this. We’ll slowly introduce more things where modes are still supported in one way or another, but they’re no longer first-class and will be deprecated in 3.0.

drwpow: Or I guess even this would be more straightforward

```diff
- transform(token, mode) {
+ transform(token, { mode }) {
    if (mode === "dark") {
```

drwpow:
I’m in favor of this change! We can make this breaking, so long as we add tests

mrginglymus:
@drwpow before I start adding tests, could I check that this implementation matches your comments above?

Currently the transform doesn't accept parameters that describe the current mode/input; we could add `input: Record<string, string>` to the options?

drwpow:
> @drwpow before I start adding tests, could I check that this implementation matches your comments above?
> 
> Currently the transform doesn't accept parameters that describe the current mode/input; we could add `input: Record<string, string>` to the options?

Oh I see. Yes let’s add that.

Minor point on terminology—we use “permutation” when it’s the complete input object with all defaults filled in. “Input” usually only refers to what the user passed in (so `{}` is a valid input that would yield the `{ "theme": "dark", "platform": "mobile" }` permutation if those modifiers had defaults set). So I think `permutation: Record<string, string>` should be added to the transform options (pretty much everywhere, even in the parser)

mrginglymus:
gotcha, makes sense. Will add that (and deal with all the horrors that will precipitate in tests :D)

mrginglymus:
dang, added to wrong branch 🥲

drwpow:
This is looking great! I see the tests have been updated with the correct typing now, and it’s added to the docs. I think only thing missing is adding some small test that tests that the transform is getting the correct values.

This is something I can help with but I’d love it if you could at least set up the expected value in tests (i.e. even if the test fails, I think it’d be better if you wrote the assertion so I don’t misinterpret a tiny detail of what you’re shooting for). Once we have a basic test in place let’s merge & ship

mrginglymus:
Great, I'll put some in first thing next week.

mrginglymus:
@drwpow done! pretty basic stuff to just check that everything is plumbed though, but includes a test for the user-land feature of being able to pre-resolve certain tokens at build time and exclude them from the output

packages/plugin-css/src/transform.ts
```diff
@@ -68,7 +68,8 @@ export default function transformCSS({
             color: { legacyHex },
             permutation: input,
           };
-          const value = customTransform?.(token, options) ?? transformCSSValue(token, options);
+          const value =
+            p.transform?.(token, options) ?? customTransform?.(token, options) ?? transformCSSValue(token, options);
```
mrginglymus, line 70: whoops , must have lost this in a conflict resolution. hurray for tests

PR #652, opened 2026-02-04
Replace Culori with Color.js

drwpow:
## Changes

Replaces Culori with [Color.js](https://colorjs.io). This results in:

- Improved performance when parsing colors
- Less glue code between DTCG and CSS specs (since it’s by the CSS Module 4 authors!)
- Better automatic color management (less finicky gamut clamping)
- Much better CSS output (see snapshot diff 👀)—GitHub Primer is now 10% lighter!

## How to Review

- Tests & snapshots updated, see diffs


packages/plugin-css/test/fixtures/color-scheme/want.css
```diff
@@ -4,17 +4,17 @@
 
 :root {
   color-scheme: light dark;
-  --foo: oklch(0.6334 0.2011 253.922);
+  --foo: oklch(63.34% 0.2011 253.9);
```
drwpow, line 7: This is pretty much all the diff—rounding errors and overall more compact colors, while still retaining their accuracy.

PR #661, opened 2026-02-07
Add Import from Figma to sidebar

drwpow:
## Changes

Adds `tz import` guide to the 2.0 docs

## How to Review

N/A

### DIFF
--- a/packages/cli/src/import/figma/index.ts
+++ b/packages/cli/src/import/figma/index.ts
@@ -33,8 +33,8 @@ export async function importFromFigma({
   unpublished,
   skipStyles,
   skipVariables,
-  fontFamilyNames,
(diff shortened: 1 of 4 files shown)


packages/cli/src/import/figma/index.ts
```diff
@@ -33,8 +33,8 @@ export async function importFromFigma({
   unpublished,
   skipStyles,
   skipVariables,
-  fontFamilyNames,
-  fontWeightNames,
+  fontFamilyNames = '/fontFamily$',
+  fontWeightNames = '/fontWeight$',
```
drwpow, line 37: Change I realized while writing the docs: allow folks to disable RegEx matching if they so desire

PR #664, opened 2026-02-10
feat: Upgrade plugin-tailwind to use Resolvers

drwpow:
## Changes

- Adds some minor breaking changes but addresses some comments raised in the Discord
- While investigating this, also added some parser optimizations that seem to speed things up by quite a lot
    - The first optimization was throwing away `merge-anything` in resolver flattening because it’s trying to be immutable (we don’t need immutability; we’re just throwing it away anyway) and handle more usecases than simple JSON
    - The 2nd was removing thousands of `wildcard-match` calls that were happening on every `getTransform()`. These calls are now being cached in memory so now only the cost of new, unique matchers are paid, lazily at the exact moment they’re needed
    - I don’t have hard benchmarks but the `transform()` step of the plugins sped up by about 25% or something silly
- Added a minor internal change, which was that tokens returned by `@terrazzo/parser` now have some properties as [getters](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get). I’m not a fan of getters in general because they can add unnecessary overhead and performance issues if used too liberally. But here they’re being used sparingly in a way that shouldn’t affect performance, but safeguard against some very hard-to-reverse correctness bugs
    - A little more detail: it has to do with the way thousands upon thousands of objects are loaded into memory whenever tokens are scanned and parsed. As we are resolving and flattening all this JSON, we do have to mutate _something_ because duplicating everything in memory is way too expensive (it can be MBs of JSON). But when/how we mutate can lead to some nasty bugs, too, if the code isn’t internally-consistent with which objects it mutates.
    - Getters throw a nice error if we try and mutate what is supposed to be a copy of another object. So for now (this will change in 3.0), the `token.mode`s are the source of truth, and `token.$value` is now just a getter for `token.mode['.'].$value`. For most users and plugins, this won’t cause any issue.

## How to Review

- Tests shoiuld pass


packages/parser/src/resolver/load.ts
```diff
@@ -135,7 +134,7 @@ export function createResolver(
         switch (item.type) {
           case 'set': {
             for (const s of item.sources) {
-              tokensRaw = merge(tokensRaw, s) as TokenNormalizedSet;
+              destructiveMerge(tokensRaw, s);
```
drwpow, line 137: Saw a significant speedup on this change. It’s a lot more “dangerous” because it requires the first object passed in receive all mutations and doesn’t create new objects in memory. but it still handles everything tokens JSON need it to

packages/plugin-css/test/fixtures/ds-github-primer/want.css
```diff
@@ -5843,7 +5843,7 @@
   --avatar-border-color: var(--border-color-translucent);
   --avatar-shadow: 0 0 0 2px var(--base-color-neutral-0);
   --base-color-black: rgb(0.39216% 1.5686% 3.5294%);
-  --base-color-neutral-13: var(--base-color-black);
+  --base-color-neutral-13: var(--base-color-white);
```
drwpow, line 5846: Note: all the CSS var diffs have been meticulously reviewed for errors. In fact they were reviewed so closely, I realized there was a bug in the upstream `dtcg-examples` package in GitHub Primer. These updated values are correct, and are unrelated to code changes here. But were a red herring making sure the core output hadn’t changed

PR #671, opened 2026-02-21
fix: better resolver support for plugin-sass

drwpow:
## Changes

Backwards-compatible change: allow plugin-sass to automatically take advantage of plugin-css’ resolver permutations without any changes. 

Also adds a small change to `@terrazzo/parser`, where if resolver transforms are specified, we try and duplicate one permutation into the default `mode: '.'` space. This is a little bit of a hack, but the goal is that legacy plugins aren’t outright broken (perhaps in some situations they get an unexpected permutation, but the fix is by upgrading all tokens to be resolver-aware then no guesswork is happening).

## How to Review


packages/plugin-css-in-js/test/index.test.ts
```diff
@@ -6,7 +6,7 @@ import { describe, expect, it } from 'vitest';
 describe('plugin-css-in-js', () => {
   it('Figma SDS', async () => {
     const cwd = new URL('./fixtures/figma-sds/', import.meta.url);
-    await execa('../../../../cli/bin/cli.js', ['build'], { cwd, stdout: 'inherit' });
```
drwpow, line 9: Note: `stout: 'inherit'` is good for debugging but it actually has a noticeable slowdown on CI runs if too much is being logged

packages/plugin-css/src/transform.ts
```diff
@@ -43,8 +43,18 @@ export default function transformCSS({
   const exclude = userExclude ? cachedMatcher.tokenIDMatch(userExclude) : () => false;
 
   // permutations
-  if (permutations?.length) {
-    for (const p of permutations) {
+  if (Array.isArray(permutations)) {
```
drwpow, line 46: These changes to plugin-css are backwards-compatible and don’t change existing behavior. As stated in PR notes, this just downgrades one permutation (the “default,” as best we can figure) into `mode: '.'`. That way legacy plugins can pick up on it.

packages/plugin-sass/src/build.ts
```diff
@@ -59,42 +59,39 @@ export default async function build({ getTransforms, options }: BuildParams): Pr
...
-      tokenValuesMap.set(tokenId, new CssVarReferenceSassToken(tokenName));
+      tokenValuesMap.set(token.id, new CssVarReferenceSassToken(tokenName));
     }
   }
   root.appendVariableDefinition('$__token-values', tokenValuesMap);
   root.appendBlankLine();
   const tokenTypographyMixinsMap = root.createMap();
 
-  for (const token of getTransforms({ format: 'css', id: '*', mode: '.', $type: 'typography' })) {
+  for (const token of getTransforms({ format: 'css', $type: 'typography' })) {
```
drwpow, line 87: The only change in `plugin-sass` is being more “ambiguous” with the query. That way it accepts whatever `plugin-css` put out, be it from resolver output or legacy modes

PR #673, opened 2026-02-21
chore: cleanup

drwpow:
## Changes

Minor test fixture cleanup:

- Formats some test fixture configs
- Simplifies some tests
- Minor types cleanup

## How to Review

- Tests should pass

### DIFF
--- a/packages/plugin-css/test/js.test.ts
+++ b/packages/plugin-css/test/js.test.ts
@@ -3,6 +3,7 @@ import { createRequire } from 'node:module';
 import { fileURLToPath, pathToFileURL } from 'node:url';
 import { build, defineConfig, parse } from '@terrazzo/parser';
 import { makeCSSVar } from '@terrazzo/token-tools/css';
+import { execaNode } from 'execa';
 import { describe, expect, it } from 'vitest';
 import css, { type CSSPluginOptions, type Permutation } from '../src/index.js';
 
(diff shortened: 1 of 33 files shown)


packages/plugin-css/test/js.test.ts
```diff
@@ -108,38 +109,25 @@ describe('Node.js API', () => {
     });
   });
 
-  describe('custom transforms', async () => {
+  describe('custom transforms', () => {
     it.each([
-      ['resolve-primitives', (await import('./fixtures/transforms-resolve-primitives/terrazzo.config.js')).default],
```
drwpow, line 113: It’s not an overt problem, but if possible it’s nice to avoid polluting the current module scope with individual test fixtures (you never know if side effects could creep in)

packages/token-tools/src/css/index.ts
```diff
@@ -36,59 +36,29 @@ export * from './stroke-style.js';
...
+  transition: transformTransition,
+  typography: transformTypography,
+};
+
 /** Main CSS Transform */
 export function transformCSSValue<T extends TokenNormalized = TokenNormalized>(
   token: T,
   options: TransformCSSValueOptions,
 ) {
-  switch (token.$type) {
```
drwpow, line 44: This is an entirely-unnecessary change, purely for my own entertainment. But if a switch statement like this is only serving the purpose of an object map, the latter will be faster.

Will it move the needle on anything? No. But stylistically I think it’s cleaner.

PR #688, opened 2026-03-12
Feature request: Partial resolver application

mrginglymus:
## Changes

Adds an option to `apply` which allows users to focus only on tokens they know they need. 

The use case for this is when outputting only a subset of tokens the resolver will (re)apply _all_ sets and modifiers.

For instance, if you are outputting tokens for light/dark modes, there is no need to apply inputs to your typography tokens.

```json
{
  "sets": {
    "primitives": [...],
    "typography": [...],
  },
  "modifiers": {
    "mode": {
      "contexts": {
        "light": [{...}],
        "dark": [{...}],
      }
    }
  }
}
```

```ts
const tokens = resolve.apply({ mode: 'light' }, { sets: ['primitives'], modifiers: ['mode'] });
```

This is definitely advanced usage, and requires manual fettling of the options to ensure everything comes through correctly, but can offer reasonable performance improvements - I'm seeing my builds locally drop from ~12.5s to ~8s.

### DIFF
--- a/packages/parser/src/lib/resolver-utils.ts
+++ b/packages/parser/src/lib/resolver-utils.ts
@@ -1,3 +1,4 @@
+import type { ResolverApplicationOptions } from '../types.js';
 import { alphaComparator } from './array.js';
 
 /**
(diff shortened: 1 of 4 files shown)


drwpow:
Ooh this would be useful! While yes, some advanced usage may be unique to certain cases, but this is pretty universally applicable. There are times like you mentioned where you know you can skip doing work, and you don’t need to recompute things you don’t need. I think this would be a great addition, allowing the resolver to be more surgical in what it’s computing 🙂.

I like the API you suggested too! No notes

mrginglymus:
@drwpow how should we deal with caching? Update `getPermutationID` to include options, or skip caching when using a filtering?

Arguably we could put some clever logic to say "if you're requesting the same permutation with a strict subset of an existing cached resolver then we can return that cached resovler", though I think that's probably not right as we'd be returning a resolver that woudl contain more tokens than you might expect.

drwpow:
Great question! Just thinking out loud:

How the full resolver works is, to make a cache key we a) take the input, b) fill in modifier defaults, c) sort keys alphabetically, then d) call `JSON.stringify()` to produce a deterministic string that makes sure that the same cache applies to different inputs if the end result is the same.

Here we could probably take a slightly-different strategy: DON’T fill in modifier defaults, and append `"sets":["a","b","c"]` to the cache key (whatever was specified). We‘d have more cache misses if there was any minor difference between partial resolvers, but I think that’s OK? The way I see it, any attempt at caching, even if it’s imperfect, is what matters.

Sidenote: if we add the literal key `"sets"`, this may cause issues if someone had a modifier named `"sets"`, which sounds silly but I think it’s actually not unreasonable. Maybe we could name it `__sets__` or something unlikely to cause collision (not `$sets` since I don’t want there to be confusion with official syntax or JSON Schema)

PR #691, opened 2026-03-16
fix(plugin-css): improve support for multiple outputs

jackholden:
## Changes
[Issue explainer](https://github.com/terrazzoapp/terrazzo/issues/687)

This PR fixes an issue where the CSS plugin was generating stale descriptions when multiple plugin instances are used to generate multiple outputs.

We now store the permutation-specific `$description` in `meta['plugin-css'].description` during `setTransform()` and have a new `getDescription()` helper which returns the metadata whilst maintaining existing backward compatibility.

Test fixtures were added and pass :)

### DIFF
--- a/packages/plugin-css/src/transform.ts
+++ b/packages/plugin-css/src/transform.ts
@@ -48,7 +48,6 @@ export default function transformCSS({
     // also duplicate the token in the global space. This plays nicer with
     // plugins that haven’t upgraded to resolvers yet.
     // If there is no “default” permutation, then take the first one
-    // let defaultPermutationI = permutations.findIndex((p) => !Object.keys(p).length);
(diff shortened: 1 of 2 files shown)


packages/plugin-css/src/transform.ts
```diff
@@ -48,14 +48,17 @@ export default function transformCSS({
     // also duplicate the token in the global space. This plays nicer with
     // plugins that haven’t upgraded to resolvers yet.
     // If there is no “default” permutation, then take the first one
-    let defaultPermutationI = permutations.findIndex((p) => !Object.keys(p).length);
+    // let defaultPermutationI = permutations.findIndex((p) => !Object.keys(p).length);
```
drwpow: We can just delete this line

Issue #704, opened 2026-03-22
TokenNormalizedSet missing from generated .d.ts imports


suikkee:
### Versions

- @terrazzo/plugin-js: 2.0.0

### Node.js version

v24.14.0

### OS + version

Ubuntu 24.04.4 LTS

### Description

 The generated `.d.ts` file references `TokenNormalizedSet` in the `PERMUTATIONS` type declaration:

```ts
export const PERMUTATIONS: Record<string, TokenNormalizedSet>;
```

However, TokenNormalizedSet was never included in the import type { ... } from
"@terrazzo/parser" block. This causes a TypeScript compilation error when
consuming the generated types.
 

### Reproduction

terrazzo.config.ts

```ts
import { defineConfig } from "@terrazzo/cli";
import css from "@terrazzo/plugin-css";
import tailwind from "@terrazzo/plugin-tailwind";
import js from "@terrazzo/plugin-js";

export default defineConfig({
  tokens: [
    "../../design/tokens/tokens.resolver.json",
  ],
  lint: {
    rules: {
      "core/consistent-naming": "off",
    },
  },
  outDir: "./src/tokens/",
  plugins: [
    css({
      skipBuild: true,
      permutations: [
        {
          input: { theme: "light" },
          prepare: (contents) => `@media (prefers-color-scheme: light) { ${contents} }`,
        },
        {
          input: { theme: "dark" },
          prepare: (contents) => `@media (prefers-color-scheme: dark) { ${contents} }`,
        },
      ],
    }),
    tailwind({
      template: new URL("tailwind.template.css", import.meta.url).pathname,
      filename: "tailwind-theme.css",
      theme: {
        theme: "light",
        color: ["semantic.color.**.*"],
      }
    }),
    js({
      filename: "tokens.js",
    }),
  ],
});

```

```shell
tz build
```

### Expected result

The generated `.d.ts` file should include `TokenNormalizedSet` in its imports:

  ```ts
  import type {
    TokenNormalizedSet,
    ColorTokenNormalized,
    DimensionTokenNormalized,
    // ...
  } from "@terrazzo/parser";

### Extra

- [x] I’m willing to open a PR (see [CONTRIBUTING.md](https://github.com/terrazzoapp/terrazzo/blob/main/CONTRIBUTING.md))

suikkee:
Sorry for opening this PR before discussing it in an issue. 
[...]

PR #705, opened 2026-03-22
feat: Add BASE_IMPORT_TYPES for consistent type imports in generated d.ts files

suikkee:
## Changes

#704

### Cause

The import generation logic in buildDTS only iterated over token types present in  TYPE_MAP, which maps individual token $type values (e.g. color, dimension) to their normalized type names. TokenNormalizedSet is not a per-token type but a structural type used for the permutation record, so it was never added to the imports.

### Fix

Introduce a BASE_IMPORT_TYPES constant in lib.ts for types that must always be present in the generated .d.ts imports, regardless of which token types are in use. The buildDTS function now emits these base types alongside the dynamic per-token types.

## How to Review

1. Run `pnpm test` in `packages/plugin-js` — the fixture snapshots should reflect `TokenNormalizedSet` in the import block.
2. The main change is in `build.ts` (import generation) and `lib.ts` (new `BASE_IMPORT_TYPES` constant). Both are small and localized.
3. I'd appreciate feedback on whether `BASE_IMPORT_TYPES` is the right abstraction, or if you'd prefer a different approach.


drwpow:
This is great, thank you! Would love a `patch` changeset added please (see comment)! Happy to approve & merge once that’s in

suikkee:
> This is great, thank you! Would love a `patch` changeset added please (see comment)! Happy to approve & merge once that’s in

Just added the patch changeset. Ready for another look!

drwpow (approved):

PR #706, opened 2026-03-24
Plugin type tweaks

mrginglymus:
## Changes

1. Marks get/set transforms as having `this: void`.
2. Allow non-async transform/build/buildEnd hooks

### DIFF
--- a/packages/parser/src/types.ts
+++ b/packages/parser/src/types.ts
@@ -328,9 +328,9 @@ export interface Plugin {
   enforce?: 'pre' | 'post';
   /** Throw lint errors/warnings */
   lint?(): Record<string, LintRule<any, any, any>>;
-  transform?(options: TransformHookOptions): Promise<void>;
-  build?(options: BuildHookOptions): Promise<void>;
-  buildEnd?(options: BuildEndHookOptions): Promise<void>;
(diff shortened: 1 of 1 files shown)


drwpow:
Sorry, I think this PR was opened while I was away and this slipped off my radar.

I’m on board with improving the overall types but could you provide a little more detail on the error this addresses? Does this occur when declaring plugins in `.ts` format? With which TS version? etc. etc.

I‘d just like to test this to make sure these changes don’t cause other TS issues for other folks, is all

mrginglymus:
Sorry for the lack of detail! Without this, accessing the `setTransform` method in a custom plugin causes a `@typescript-eslint/unbound-method` error:

```ts
export default defineConfig({
  plugins: [
    {
      transform({ resolver, tokens, setTransform }) {
```

with the suggestion

> If a function does not access `this`, it can be annotated with `this: void`.

The adding of `void | Promise<void>` is to allow for methods that don't have await calls without having to suppress `@typescript-eslint/require-await` - although arguably `@typescript-eslint` should be able to detect that the function is being passed to something that only accepts awaitable functions.

drwpow (approved):
Ah OK, works for me. Thanks!

PR #708, opened 2026-03-24
Set transform tokensset

mrginglymus:
- fixes #707

## Changes

ensures that `setTransform` looks up tokens on the applied input rather than the default input.

### DIFF
--- a/packages/parser/src/build/index.ts
+++ b/packages/parser/src/build/index.ts
@@ -141,11 +141,6 @@ export default async function build(
             });
             return;
           }
-          const tokens = resolver.apply(params.input ?? {});
-          const token = tokens[id]!;
-          if (!token) {
-            logger.error({ group: 'plugin', label: plugin.name, message: `No token "${id}"` });
-          }
           const isLegacyModes =
             params.input && Object.keys(params.input).length === 1 && 'tzMode' in params.input;
           const permutationID =
@@ -162,6 +157,14 @@ export default async function build(
             params: { ...(params as any), value: cleanValue },
             pluginName: plugin.name,
           });
+
+          // If we are in legacy mode, use the provided tokens. If we are in resolver mode,
+          // use the current input to ensure we capture tokens that might not have been declared under
+          // the initial default input
+          const token =  (mode ? tokens : resolver.apply(params.input ?? {}))[id]!;
+          if (!token) {
+            logger.error({ group: 'plugin', label: plugin.name, message: `No token "${id}"` });
+          }
 
           // upsert
           if (!formats[params.format]) {
--- a/packages/parser/src/build/index.ts
+++ b/packages/parser/src/build/index.ts
@@ -141,6 +141,7 @@ export default async function build(
             });
             return;
           }
+          const tokens = resolver.apply(params.input ?? {});
           const token = tokens[id]!;
           if (!token) {
             logger.error({ group: 'plugin', label: plugin.name, message: `No token "${id}"` });
--- a/packages/parser/test/plugin.test.ts
+++ b/packages/parser/test/plugin.test.ts
@@ -47,6 +47,62 @@ describe('Plugin API', () => {
       }
     });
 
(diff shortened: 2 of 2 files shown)


mrginglymus:
Hmm, this has had a wider impact that anticipated.

Some of the changes just look like re-ordering, but there are also some dropped tokens. Is this a caching issue again?

drwpow:
> Hmm, this has had a wider impact that anticipated.
> 
> Some of the changes just look like re-ordering, but there are also some dropped tokens. Is this a caching issue again?

Yeah possibly. I think as an approach we DON’T want to call `resolver.apply(params.input ?? {})` here because this would automatically break many valid resolvers without defaults in modifiers. We may have to do something earlier, where as we’re parsing files, we just build a tokens list of all IDs as we’re crawling.

There’s a [refMap](https://github.com/terrazzoapp/terrazzo/blob/main/packages/parser/src/parse/token.ts#L252) data structure where we reference which tokens should alias to what. I forget if we always add all tokens to that structure, but it wouldn’t hurt to simply add all tokens as we encounter them there (even if the alias is nothing), and then we’d at least have a complete list of every ID encountered.

Anyway doing something more “active” like this would be better performance-wise and all.

But also, yes, it does seem like there may be a caching issue, too 🤔. Maybe we can figure that out separately if we can get around calling `.apply()` without the user requesting it.

drwpow:
Sorry for the delay in following up on this. There were some other caching bugs that have been resolved, hence some of the conflicts. This may have been affected by that! If we could rebase here & reasses the changes, we may be in a good place

mrginglymus:
@drwpow-figma done! looks like it's just re-ordered things now?

mrginglymus:
o no....node version sensitive ordering bugs?

drwpow:
🤔 Hm. Yeah I wonder if we should have just kept alphabetization, after all. Especially when scanning tokens, it almost seems non-deterministic ending up with a random ordering each time. I don’t remember who requested tokens should just appear in order of appearance, but I am thinking that was a mistake (I didn’t have an opinion at the time). 

I’m starting to feel now that just alphabetizing always, and enforcing a predictable order is better. And if folks want to have their tokens appear in a certain order, just… enforce that with naming because that’s the whole point of naming (e.g. [category → variant → state](https://thedesignsystem.guide/design-tokens) or whatever convention folks follow for taxonomy).

mrginglymus:
I think for tests it makes sense to turn it on. Perhaps the default should be to sort, but an escape hatch can be added to allow emitting in declaration order (ie, skip sorting). I guess the problem is that it's hard to pinpoint what declaration order actually is when you're scanning multiple files like this.

drwpow:
OK merged #769 which just alphabetizes CSS. If you rebase off that one, and regenerate snapshots, hopefully we should only see meaningful diffs here!

drwpow:
Hm I’m seeing an error in the tests. We need this to be a backwards-compatible change, so the existing tests should pass with this PR. But happy to approve & merge once tests are passing again.

mrginglymus:
There is also still...weirdness in ordering. Will poke around to find what's gone wrong.

mrginglymus:
OK @drwpow the problem I'm hitting now is that the initial internal input application applies defaults based on either an explicit default, or an assumed default (based on 'first context'). However, subsequent external calls to `apply` only use the explicit defaults. 

This means that usages of the resolver that relied on the 'first context' default will now explode as the new call to `apply` is no longer complete.

How would we feel about unifying the behaviour to always use implicit defaults?

mrginglymus:
You know...the more I look at this the more I think this is just the wrong thing to do. Certainyl doing it in `setTransform` doesn't seem to make sense, as that is only called with the ids of tokens that were in the original tokensset...also...it's a lot easier to just say "all your contexts must be homogenous".

mrginglymus:
(that being said, the inconsistency of the first application using implicit + explicit defaults and subsequent applications only using explicit defaults _has_ caught me out a couple of times)

PR #715, opened 2026-04-09
Fix Figma hidden local alias resolution and emitted dependency closure

heddendorp:
## Changes

Figma import was dropping `hiddenFromPublishing` local variables before alias resolution, so aliases to hidden local targets were misclassified as remote and omitted from the generated resolver. This PR now fixes both parts of the issue: hidden locals remain available for alias lookup, and the emitted resolver includes the local alias dependency closure needed to keep the output buildable.

- **Importer flow**
  - Keep the full local variable set in `allVariables`
  - Use `allVariables` as the alias-resolution source of truth
  - Seed `finalVariables` from the published endpoint in published mode, or from non-hidden locals in unpublished mode
  - Expand the emitted variable set by walking local `VARIABLE_ALIAS` references and recursively adding any required local targets from `allVariables`

- **Behavioral impact**
  - Published variables can now resolve aliases to hidden local variables in the same file
  - `--unpublished` still excludes hidden variables from the initial emitted root set, but now includes any hidden local alias dependencies required to make the resolver self-contained
  - The generated resolver no longer contains dangling local alias references for hidden targets, which addresses the downstream build failure
  - Truly remote variables are still counted and warned about when they cannot be accessed

- **Coverage**
  - Add focused tests for hidden local alias dependency closure, including recursive local alias chains
  - Update Figma import snapshots to reflect the self-contained emitted resolver output

```ts
for (const id of Object.keys(local.meta.variables)) {
  allVariables[id] = local.meta.variables[id]!;
}

const pendingIDs = Object.keys(finalVariables);
for (let i = 0; i < pendingIDs.length; i++) {
  const variable = finalVariables[pendingIDs[i]!];
  for (const value of Object.values(variable.valuesByMode)) {
    const aliasID = getAliasID(value);
    if (aliasID && allVariables[aliasID] && !(aliasID in finalVariables)) {
      finalVariables[aliasID] = allVariables[aliasID]!;
      pendingIDs.push(aliasID);
    }
  }
}
```

## How to Review

Review `packages/cli/src/import/figma/variables.ts` first.

- Confirm the separation of concerns:
  - `allVariables` is the complete local lookup map
  - `finalVariables` is the emitted root set
  - emitted output now includes the recursive local alias dependency closure required by those roots
- Confirm published mode still starts from published variables, while unpublished mode still starts from visible local variables
- Check `packages/cli/test/figma-variables.test.ts` for the new assertions that hidden local alias targets are emitted when required for buildable output
- Skim the updated Figma snapshots to verify the additional emitted tokens are local alias dependencies, not unrelated output changes


drwpow (approved):
Thanks for adding this! To be honest I’m not sure about all the changes in the snapshot files, but I do see more tests updated, and existing ones passing, so even if we accidentally included too many variables, that’s fine! Like you pointed out, missing variables are broken either way, so this seems like a good fix

PR #719, opened 2026-04-10
fix plugin css mode aliases

drwpow:
## Changes

Fixes a bug in `plugin-css` where legacy `modeSelectors` would sometimes accidentally grab the alias from the default mode rather than the requested mode. It was flaky, and depended on the original `setTransform()` order.

## How to Review

- Tests updated


packages/plugin-css/test/fixtures/mode-type-color/index.want.css
```diff
@@ -122,15 +112,10 @@
     --ds-color-blue-3: rgb(34.51% 65.098% 100%);
     --ds-color-blue-4: rgb(21.961% 54.51% 99.216%);
     --ds-color-blue-5: rgb(12.157% 43.529% 92.157%);
+    --ds-color-text-action: var(--ds-color-blue-5);
```
drwpow, line 115: This was easy to miss, but the bug was apparent here, just hidden amongst the noise.

PR #726, opened 2026-04-14
Fix token aliases

drwpow:
## Changes

Fixes another bug related to token aliasing. In this scenario, say you have a primitive token with multiple modes, and your semantic tokens simply reference them. These would accidentally get dropped in the hoisting.

#719 didn’t actually fix the problem, it actually just put a bandaid over it. This PR should fix the problem more holistically.

## How to Review

- Tests updated

### DIFF
--- a/packages/plugin-css/test/fixtures/mode-type-color/index.want.css
+++ b/packages/plugin-css/test/fixtures/mode-type-color/index.want.css
@@ -42,6 +42,11 @@
 
 @media (prefers-color-scheme: light) {
   :root {
+    --ds-color-alias-1: var(--ds-color-blue-6);
+    --ds-color-alias-2: var(--ds-color-alias-1);
+    --ds-color-alias-3: var(--ds-color-alias-2);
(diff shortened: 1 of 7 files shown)


packages/plugin-css/test/fixtures/mode-type-color/index.want.css
```diff
@@ -42,6 +42,11 @@
 
 @media (prefers-color-scheme: light) {
   :root {
+    --ds-color-alias-1: var(--ds-color-blue-6);
+    --ds-color-alias-2: var(--ds-color-alias-1);
+    --ds-color-alias-3: var(--ds-color-alias-2);
+    --ds-color-alias-4: var(--ds-color-alias-3);
+    --ds-color-alias-5: var(--ds-color-alias-4);
```
drwpow, line 49: This should have been the signal it wasn’t working—these tokens are meant to test hoisting! With snapshot tests it’s hard, because the upside is they test dozens of cases, but the downside is they test dozens of cases. In some instances it’s not even maintainable to break them out into unit tests.

PR #735, opened 2026-04-16
fix: normalize token IDs regardless of alphabetize setting

### LINKED ISSUE #734: Parsed token IDs differ depending on `alphabetize` option
### Versions

- `@terrazzo/parser@2.0.3`

### Node.js version

20.19.6

### OS + version

macOS 26.4.1

### Description

The `tokens` result of `@terrazzo/parser` will switch between JSON Pointer format (`#/color/red`) when `alphabetize: false`, and dot notation (`color.red`) when `alphabetize: true`.

This behavior likely changed in the 2.x version, as it's an issue I'm encountering in trying to update our project, which has a few custom plugins that operate on the tokens.

Issue is likely in here, where `alphabetize: false` short-circuits early, but the logic that follows for `alphabetize: true` manipulates the tokens shape differently from what would be returned early:

https://github.com/terrazzoapp/terrazzo/blob/fefd4b149d03ca2199eaddc7b197e4e081169e7f/packages/parser/src/parse/process.ts#L256-L272

### Reproduction

(Caveating that this reproducible case are AI-generated snippets)

`tokens.json`:

```json
{
  "size": {
    "$type": "dimension",
    "sm": {
      "$value": { "value": 4, "unit": "px" }
    },
    "md": {
      "$value": { "value": 8, "unit": "px" }
    },
    "base": {
      "$value": "{size.sm}"
    }
  }
}

```

`test.mjs`:

```ts
import { defineConfig, parse } from "@terrazzo/parser";
import { readFileSync } from "node:fs";

const src = readFileSync(new URL("./tokens.json", import.meta.url), "utf8");

for (const alphabetize of [true, false]) {
  const config = defineConfig(
    { tokens: ["./tokens.json"], outDir: "./out", plugins: [], alphabetize },
    { cwd: new URL("./", import.meta.url) }
  );

  const { tokens } = await parse(
    [{ filename: config.tokens[0], src }],
    { config, skipLint: true }
  );

  console.log(`\nalphabetize: ${alphabetize}`);
  for (const [id, token] of Object.entries(tokens)) {
    console.log(`  id: ${id}  aliasOf: ${token.aliasOf ?? "(none)"}`);
  }
}
```

Output:

```
alphabetize: true
  id: size.base  aliasOf: size.sm
  id: size.md  aliasOf: (none)
  id: size.sm  aliasOf: (none)

alphabetize: false
  id: #/size/sm  aliasOf: (none)
  id: #/size/md  aliasOf: (none)
  id: #/size/base  aliasOf: size.sm
```

### Expected result

`id` should be consistent regardless of `alphabetize` option

### Extra

impsislegobatman:
## Summary

Fixes #734

When `alphabetize` is set to `false`, the early return in `process.ts` (line 258) skipped the `refToTokenID()` conversion, causing token IDs to remain in JSON Pointer format (`#/size/sm`) instead of dot notation (`size.sm`).

This change ensures `refToTokenID()` always runs on token IDs while only conditionally sorting them, so token ID format is consistent regardless of the `alphabetize` setting.

## Changes

- Removed the early return when `alphabetize === false`
- Always apply `refToTokenID()` conversion to normalize token IDs
- Conditionally sort with `alphaComparator` only when `alphabetize !== false`
- Conditionally sort group token arrays only when `alphabetize !== false`

## Test plan

- All 258 existing tests pass
- Verified build succeeds

### DIFF
--- a/packages/parser/src/parse/process.ts
+++ b/packages/parser/src/parse/process.ts
@@ -253,24 +253,24 @@ export function processTokens(
   }
   logger.debug({ ...entry, message: 'Normalized values', timing: performance.now() - normalizeStart });
 
-  // 6. alphabetize & filter
+  // 6. normalize IDs & optionally alphabetize
   // This can’t happen until the last step, where we’re 100% sure we’ve resolved everything.
-  if (config.alphabetize === false) {
-    return tokens;
-  }
-
   const sortStart = performance.now();
-  const tokensSorted: TokenNormalizedSet = {};
-  tokenIDs.sort(alphaComparator);
+  if (config.alphabetize !== false) {
+    tokenIDs.sort(alphaComparator);
+  }
+  const tokensNormalized: TokenNormalizedSet = {};
   for (const path of tokenIDs) {
     const id = refToTokenID(path)!;
-    tokensSorted[id] = tokens[path]!;
+    tokensNormalized[id] = tokens[path]!;
   }
-  // Sort group IDs once, too
-  for (const group of Object.values(groups)) {
-    group.tokens.sort(alphaComparator);
+  if (config.alphabetize !== false) {
+    // Sort group IDs once, too
+    for (const group of Object.values(groups)) {
+      group.tokens.sort(alphaComparator);
+    }
   }
   logger.debug({ ...entry, message: 'Sorted tokens', timing: performance.now() - sortStart });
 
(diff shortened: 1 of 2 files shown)


drwpow (approved):
Amazing, thank you! We should release this.

PR #736, opened 2026-04-16
Add `variableName` param to plugin tailwind

### LINKED ISSUE #721: `makeCSSVar` strips underscores and collapses double-dashes in CSS variable names
### Versions

- `@terrazzo/cli: 2.0.1`
- `@terrazzo/plugin-css: 2.0.1`
- `@terrazzo/plugin-tailwind: 2.0.1`
- `@terrazzo/token-tools: 2.0.1`

### Node.js version

24.7.0

### OS + version

macOS 26.0.1

### Description

`makeCSSVar` in `@terrazzo/token-tools` uses regex to split and rejoin (`CSS_VAR_RE`) that treats underscores and hyphens as word boundaries. This strips underscores and collapses double-dashes, both of which are valid in CSS custom property names. I respect that the current approach is a strong default we should preserve but it can be limiting for some token workflows that use existing libraries that need to match output to maintain compatibility.

In `@terrazzo/token-tools/dist/css.js`
```typescript
const CSS_VAR_RE = /(?:(\p{Uppercase_Letter}?[\p{Lowercase_Letter}\p{Number}]+|[\p{Uppercase_Letter}\p{Number}]+|[\u{80}-\u{10FFFF}\p{Number}]+)|.)/u;

let property = name.split(CSS_VAR_RE).filter(Boolean).join("-");
```

The `.` alternative in the regex matches `_` and `-` as single-character tokens, which get replaced when rejoined with `-`.

### Reproduction

Token names with underscores (DTCG-valid, common in design systems for variant separation):
```
primary_hover   → --primary-hover      (underscore stripped)
brand-50_alt     → --brand-50-alt        (underscore stripped)
```
Token names requiring double-dashes (Tailwind v4 convention for paired properties):
```
xs--line-height   → --xs-line-height     (double-dash collapsed)
```

This affects tokens in our design system that use underscores as variant separators like _hover, _alt, _subtle, _on-brand, _disabled, _pressed, and prevents generating Tailwind v4's `--text-*--line-height` and `--text-*--letter-spacing` paired properties.

### Expected result

Underscores and double-dashes should be optionally preserved in the output as both are valid in CSS custom property names.

### Extra

burkeshartsis:
## Changes

Adds a `variableName` option to `@terrazzo/plugin-tailwind`. The callback receives the default name and a context object (`token`, `path`, `relName`), so users can selectively override CSS variable names and fall back to the default for everything else. This matches the existing `variableName` pattern in `plugin-css`.

Closes #721

## How to Review

The test fixture (`test/fixtures/variable-name/`) covers three cases:
1. Underscore preservation (`--color-bg-primary_hover`)
2. Double-dash preservation (`--text-xs--line-height`)
3. `defaultName` fallback (`--text-font-small`) ensures no breaking changes

The Tailwind compilation test verifies all three cases survive end-to-end through the real `tailwindcss` CLI.

Docs updated in `www/src/pages/docs/integrations/tailwind.md`.


burkeshartsis:
@drwpow any chance we could look at getting this merged some time soon? I'm motivated to help any way I can.

drwpow (approved):
Great work! Great tests, too! I don’t have any feedback. Through communicating about this over the issue, and the documentation updates here, this is a great add.

Sorry this didn’t make it into this week’s release, but this will be queued up for the next one 🙂

PR #737, opened 2026-04-16
fix(parser): normalize token IDs to dot-notation when alphabetize=false

### LINKED ISSUE #734: Parsed token IDs differ depending on `alphabetize` option

armorbreak001:
## Summary

Fixes #734

## Problem

When `alphabetize: false`, `processTokens()` returned tokens with **JSON Pointer** keys (e.g. `#/color/red`) instead of **dot-notation** keys (e.g. `color.red`). This inconsistency meant that plugins and downstream code received different token ID formats depending on the alphabetize setting.

## Root Cause

The `alphabetize === false` path returned the raw `tokens` object (keyed by JSON Pointer `jsonID`) without normalizing IDs through `refToTokenID()`. The `alphabetize: true` path correctly converted keys to dot notation, but the false path skipped this step.

## Fix

- Extract token ID normalization into a shared step that runs **before** the alphabetize branch
- Both paths now return tokens with consistent dot-notation keys
- Added test coverage for `alphabetize: false` output format

## Changes

- `packages/parser/src/parse/process.ts`: Normalize token IDs via `refToTokenID()` before the alphabetize check
- `packages/parser/test/parse.test.ts`: New test verifying dot-notation output when `alphabetize: false`

## Test Plan

- [x] All existing tests pass (259 pass, 1 skip)
- [x] New test verifies `alphabetize: false` produces same key format as `alphabetize: true`

### DIFF
--- a/packages/parser/src/parse/process.ts
+++ b/packages/parser/src/parse/process.ts
@@ -256,7 +256,15 @@ export function processTokens(
   // 6. alphabetize & filter
   // This can’t happen until the last step, where we’re 100% sure we’ve resolved everything.
   if (config.alphabetize === false) {
-    return tokens;
+    // Normalize token IDs to dot-notation for consistency with alphabetize=true
+    const tokensNormalized: TokenNormalizedSet = {};
+    for (const path of tokenIDs) {
+      const id = refToTokenID(path)!;
+      if (id) {
+        tokensNormalized[id] = tokens[path]!;
+      }
+    }
+    return tokensNormalized;
   }
 
   const sortStart = performance.now();
--- a/packages/parser/test/parse.test.ts
+++ b/packages/parser/test/parse.test.ts
@@ -1190,4 +1190,20 @@ describe('Transform', () => {
       hex: '#0088ff',
     });
   });
+
+
+  it('alphabetize=false produces dot-notation token IDs (same as alphabetize=true)', async () => {
+    const src = {
+      color: {
+        red: { $type: 'color', $value: { colorSpace: 'srgb', components: [1, 0, 0], alpha: 1 } },
+        blue: { $type: 'color', $value: { colorSpace: 'srgb', components: [0, 0, 1], alpha: 1 } },
+      },
+    };
+    const config = defineConfig({ alphabetize: false }, { cwd });
+    const { tokens } = await parse([{ filename: DEFAULT_FILENAME, src }], { config });
+    // Token IDs should be in dot-notation, NOT JSON Pointer format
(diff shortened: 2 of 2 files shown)


drwpow:
Thank you! Sorry I didn’t realize #735 also covers the same fix, but that PR takes the additional step of making sure the groups are correctly ordered, too.

PR #738, opened 2026-04-16
Add omitTypographyShorthand & subValueVariableName options to CSS plugin

### LINKED ISSUE #720: More control over CSS plug-in's output for typography tokens
### Summary

Currently, it seems the CSS plugin will output the following CSS for a typography token (give a typography token with the id `my.typo.token`):

```css
/* Individual custom properties for the sub-values */
--my-typo-token-font-family: ...;
--my-typo-token-font-weight: ... ;
--my-typo-token-font-size: ...;
--my-typo-token-line-height: ...;
--my-typo-token-letter-spacing: ...;

/*  Custom property for the shorthand typography style */
--my-typo-token: var(--my-typo-token-font-weight) var(--my-typo-token-font-size)/var(--my-typo-token-line-height) var(--my-typo-token-font-family);
```

What I need for my design system is:

* The ability to omit the shorthand style (we have our own mechanism for combining the sub-values into a CSS classes downstream, so it's redundant for us). I don't see any option to do this, and - if I've read the code right - that behaviour is hard coded.
* The ability to customise how the sub-value names are appended. Right now, there's [a hard-coded hyphen separator](https://github.com/terrazzoapp/terrazzo/blob/3a70a7037c891d15d330425177dd5d45affcb7b2/packages/plugin-css/src/build.ts#L116). In our case, we need to use `__` instead.

Essentially, the CSS output I would like is:

```css
/* Individual custom properties for the sub-values */
--my-typo-token__font-family: ...;
--my-typo-token__font-weight: ... ;
--my-typo-token__font-size: ...;
--my-typo-token__line-height: ...;
--my-typo-token__letter-spacing: ...;

/*  Intentionally omit the shorthand custom prop */
```

I appreciate that the output I'd like is probably a niche use-case. Nevertheless, I think extending the CSS plugin to make this possible wouldn't be too hard and perhaps it'll help out some other folks too.

Having said that, I'm open to alternative approaches that would let me solve this myself. I know I could make a custom CSS plugin, but that seems overkill.

I had wondered if I could create a custom plugin that just tweaks the CSS plugin's transforms _before_ its build step runs, but I couldn't find a way of doing that, as:

- In the transforms, the typography token is still a single entry, the mapping to multiple CSS custom properties happens in the build step.
  - Although, I did wonder if I could edit the transform to convert it to a "single-value", for one of the sub-values, and then add additional transforms for the other sub-values. However, I don't know if having multiple transforms based on the same token would cause issues or not. Also, this feels a bit hacky. I'd prefer to work "with the grain" of Terrazzo and its APIs, rather than against it :-)
- I also considered building a custom build step to replace the CSS plugin's one, but again, that seems overkill when most of the time, I'm perfectly happy with the plugin's existing output.

However, if I'm mistaken or have missed some other way of achieving what I want, please do say so.

### Proposal & prior art

I propose adding a couple of extra config options to the CSS plugin:

1. To omit the output of shorthand properties.
  - This could either be a `boolean`, or perhaps a function that gets the token's `$type` (or even the whole token?) and returns a boolean (`true` to omit, `false` to keep) to allow more fine-grained control. Or both?
2. To customise how sub-values names are appended
  - Maybe a function would be best here, to give more control and be similar to the existing `variableName` option.

So, something like this perhaps:

```ts
interface CSSPluginOptions {
  // ...

  omitMultiValueShorthands?: boolean | (token? TokenNormalized) => boolean;

  subValueVariableName?: (localID: string, name: string, token: TokenNormalized) => string;
}
```

### Extra

james-nash:
## Changes

Adds new options to the CSS plug-in:

- `omitTypographyShorthand`: Which prevents a shorthand CSS variable from being output in addition to the per-sub-value ones for typography tokens.
- `subValueVariableName`: Lets you provide a function that controls how sub-values are appended to CSS variable names (currently, it's hard-coded to be hyphen-separated).
  - @drwpow had suggested extending the existing `variableName` option. However, on closer inspection, that's used during the transform phase to take a `TokenNormalized` and return a CSS variable name. The creation of per-sub-value variables doesn't occur until the _build_ phase though, so you're dealing with `TokenTransformed` at that point. I therefore felt that, making this a separate option, rather than overloading `variableName` was more appropriate. Happy to be challenged on that though!
  - Not sure how useful it'll be, but I've made it so that the current `TokenTransformed` is also passed into the `subValueVariableName` function. In theory that might allow things like varying how sub-value names are formatted based on the token's type or something like that. Happy to remove that though, if you feel it's overkill.

Closes #720 

## How to Review

The changes are fairly small and localised, so hopefully just looking at the diffs should suffice. Happy to explain or walk-through anything though. I've tried to align the existing code structure and patterns.

In addition to the actual implementation of those options, I've also:
- Added tests
- Added changesets
- Updated the CSS plugin's docs page

If I've missed anything please let me know!


drwpow (approved):
> However, on closer inspection, that's used during the transform phase to take a TokenNormalized and return a CSS variable name. The creation of per-sub-value variables doesn't occur until the build phase though, so you're dealing with TokenTransformed at that point. I therefore felt that, making this a separate option, rather than overloading variableName was more appropriate. Happy to be challenged on that though!


I can appreciate that! I agree, I don’t think this would cause issues or yield conflicts I can predict.

This looks good to merge! Sorry there’s a bit of a conflict because of another PR merging, but once that’s resolved, happy to release

james-nash:
@drwpow Thanks for reviewing my PR. I've just rebased and resolved the conflicts. This should now be good to merge whenever you're ready.

james-nash:
Just spotted the requirement for verified commits, so have force-pushed the same changes but with verified commits now.

james-nash:
Whenever the maintainers have a spare moment...

image

drwpow:
I’M SORRY JAMES I GOT BUSY. IT IS MERGED

james-nash:
No worries at all. Thanks for merging and releasing this! ❤️

PR #740, opened 2026-04-21
feat: add OAuth bearer token support for Figma import

### LINKED ISSUE #739: Figma: Add OAuth Support
### Summary

Current authentication method only supports PATs, but in most companies this is not ideal - Figma recommends developers [create an OAuth app](https://developers.figma.com/docs/rest-api/authentication/).

### Proposal & prior art

Adding support for OAuth is fairly trivial, instead of the `X-Figma-Token: XXX` header, we need `Authorization: Bearer XXX `, I propose we introduce support for a new environment variable `FIGMA_OAUTH_TOKEN` to handle this.

This also provides an opportunity to cleanup the current code where `X-Figma-Token` is manually injected into each request, instead a helper can be used to generate the headers, `FIGMA_OAUTH_TOKEN` can take priority when both are present.

### Extra

badams:
## Changes

Introduces OAuth support for Figma via the `FIGMA_OAUTH_TOKEN`

## How to Review

- fixes #739

### DIFF
--- a/.changeset/figma-oauth-support.md
+++ b/.changeset/figma-oauth-support.md
@@ -1,0 +1,5 @@
+---
+"@terrazzo/cli": minor
+---
+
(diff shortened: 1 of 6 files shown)


drwpow (approved):
Awesome. Makes perfect sense. Also, great tests!

PR #745, opened 2026-04-22
fix(plugin-css-in-js): camelCase intermediate path segments

unpunnyfuns:
## Changes

The walker wrote `node[camelCase(next)]` but read `node[next]`, crashing on dashed DTCG paths (`color.accent.bg-hover`, `number.line-height.loose`). Aligns the read with the write.

## How to Review

Added tests and fixtures.


drwpow (approved):
Great fix, thank you! 😅 I’m familiar with this code, and can definitely see this happening. Great test, too

PR #748, opened 2026-04-24
chore: Update deps

drwpow:
## Changes

Bumps deps to latest. Also updates TypeScript to 6.x, which includes some minor breaking changes in preparation for 7.x (the fast one)

## How to Review

- CI


packages/cli/tsconfig.json
```diff
@@ -1,8 +1,7 @@
 {
   "extends": "../../tsconfig.base.json",
   "compilerOptions": {
-    "baseUrl": "src",
```
drwpow, line 4: This was never meant to be here, since we don’t use absolute imports. This was just a single error that got copy-pasted too many times

packages/plugin-css-in-js/tsconfig.json
```diff
@@ -1,8 +1,8 @@
 {
   "extends": "../../tsconfig.base.json",
   "compilerOptions": {
-    "baseUrl": "src",
-    "outDir": "dist"
+    "outDir": "dist",
+    "types": ["node"]
```
drwpow, line 5: This is going to come up a lot more with `isolatedModules`, and I’m in agreement for it: more explicit types. No longer will `@types/*` just wreak havoc on a monorepo, it will have to be opted in

PR #753, opened 2026-04-27
Add hono dependency to @terrazzo/cli

mrginglymus:
## Changes

@hono/node-server has a peer dependency on hono; in this use case it feels not unreasonable for @terrazzo/cli to provide this dependency.

Alternative approaches would be making `hono` a peer dep of `@terrazzo/cli`, or making `@hono/node-server` an optional peer dep with a lazy import so that those not using `lab` don't need to install it

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -55,6 +55,7 @@
     "detect-package-manager": "^3.0.2",
     "dtcg-examples": "catalog:",
     "escodegen": "^2.1.0",
+    "hono": "^4.12.15",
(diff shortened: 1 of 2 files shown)


drwpow:
Thanks for suggesting, but for now, we’re just going to remove Hono until the Lab is ready. There are also security patches and other issues are getting flagged for, while not actually using Hono (Token Lab is my white whale, apparently, and something always prevents me from working on it). In another PR I’m going to bump dependencies and just remove Hono altogether to solve this _and_ some related issues. But thank you for the fix all the same!

PR #754, opened 2026-04-27
Do not provide listPermutations API for complex resolvers

### LINKED ISSUE #752: `listPermutations` potentially lethal - add safety checks
### Summary

Whilst `listPermutations` works fine for the classic `mode` + `theme` combo, it is catastrophic on more advanced setups. Our resolver is currently at 15 million permutations, and is likely to grow. The vast majority of these permutations affect only a handful of tokens each.

### Proposal & prior art

The `listPermutations` api is a sure-fire way to crash the process at the moment. Both of the two downstream terrazzo consuming projects I've seen use it, rendering them unusable for my project.

I'm not sure what the correct solution here is. A couple spring to mind:

```ts
listPermutations() {
  if (Object.values(validContexts).reduce((acc, contexts) => acc * contexts.length, 1)) > 100) {
    // At least the browser wont hang/node won't OOM
    throw new Error("Too many permutations!")
    // Or we could just return []
  }
}
```

```ts
listPermutations(modifiers: string[]) {
  // Require users to provide a list of modifiers to include in permutations
  return calculatePermutations(Object.entries(validContexts).filter(([name]) => modifiers.includes(name))
}
```

However, the first would require downstream callers to handle the API not working, and the second could still be abused by downstream callers to list all permutations.

A more substantial but probably safer option is to add a terrazzo extension to the DTCG spec to add an `includeInPermutations` flag to modifier definitions. If at least one of these is set, then `listPermutations` will only use those in computing permutations. 

This would allow token authors to opt only modifiers that they consider relevant to listPermutations (ie, mode and theme). However, even that leaves a lot of unanswered questions and ambiguity over the purpose of the API.

### Extra

mrginglymus:
## Changes

- fixes #752

Obviously we should provide a way of configuring the limit, but this seems like the safest way of doing it. It's a breaking change, but at least it's a _visible_ breaking change rather than invisibly throwing an error or returning `[]`.

### DIFF
--- a/packages/cli/src/shared.ts
+++ b/packages/cli/src/shared.ts
@@ -55,6 +55,7 @@ export async function loadConfig({ cmd, flags, logger }: LoadConfigOptions) {
       },
       alphabetize: false,
       ignore: { tokens: [], deprecated: false },
+      permutationLimit: 1000,
     };
     let configPath: string | undefined;
 
--- a/packages/parser/src/config.ts
+++ b/packages/parser/src/config.ts
@@ -26,6 +26,7 @@ export default function defineConfig(
   normalizePlugins({ config, logger });
   normalizeLint({ config, logger });
   normalizeIgnore({ config, logger });
+  normalizePermutationLimit({ config, logger });
 
   // 2. Start build by calling config()
   for (const plugin of config.plugins) {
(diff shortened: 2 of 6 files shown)


packages/cli/src/shared.ts
```diff
@@ -55,6 +55,7 @@ export async function loadConfig({ cmd, flags, logger }: LoadConfigOptions) {
       },
       alphabetize: false,
       ignore: { tokens: [], deprecated: false },
+      permutationLimit: 1000,
```
drwpow, line 58: 👍 this is a reasonable limit

drwpow (approved):

drwpow:
> It's a breaking change, but at least it's a _visible_ breaking change

Right, we discussed in Discord a bit, and this is sort of a “breaking in _theory_, but in practice, nothing has changed.” Because breaking changes really equate to _breaking workflows_, and this was a workflow that was not possible. So this does not warrant a major version, because no one will be affected in practice.

PR #756, opened 2026-05-06
Round CSS gradient stop positions to avoid IEEE-754 noise

unpunnyfuns:
## Changes

`transformGradient` emits each gradient stop position via raw `${100 * position}%`. IEEE-754 makes `0.55` land in CSS as `55.00000000000001%`. Switching to `${+(100 * position).toFixed(3)}%` rounds to 1/1000.

## How to Review

Two new tests in `packages/token-tools/test/css.test.ts` cover the noise-rounding case (`position: 0.55` → `55%`) and bounded precision (`position: 1/3` → `33.333%`). No fixture in `packages/*/test/fixtures/` uses a non-trivial position, so no plugin snapshots churn.

```sh
pnpm --filter @terrazzo/token-tools test
```

### DIFF
--- a/.changeset/gradient-stop-float-precision.md
+++ b/.changeset/gradient-stop-float-precision.md
@@ -1,0 +1,5 @@
+---
+"@terrazzo/token-tools": patch
+---
+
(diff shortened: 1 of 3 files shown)


drwpow (approved):
Nice fix, thank you!

PR #762, opened 2026-05-19
Allow partial application of input

mrginglymus:
## Changes

- add `resolveAliases` option to `Resolver#apply` API
- ensure inline aliasing step respects resolveAlias

Why?

This now allows you to do:

```ts
const tokens = resolver.apply({}, { sets: ['a-set'], modifiers: [], resolveAliases: false });
```

and the result will contain _only_ tokens declared in `a-set`. The same is possible for modifiers.

This is of great benefit when writing plugins that need to output only the tokens declared in a given set/modifier (e.g., in the case of a css class per context per modifier).

### DIFF
--- a/packages/parser/src/types.ts
+++ b/packages/parser/src/types.ts
@@ -351,6 +351,7 @@ export type ResolverInput = Record<string, string>;
 export type ResolverApplicationOptions = {
   sets?: string[];
   modifiers?: string[];
+  resolveAliases?: boolean;
 };
 
 export interface Resolver<
--- a/packages/parser/src/types.ts
+++ b/packages/parser/src/types.ts
@@ -349,9 +349,30 @@ export interface ReferenceObject {
 export type ResolverInput = Record<string, string>;
 
 export type ResolverApplicationOptions = {
-  sets?: string[];
-  modifiers?: string[];
+  /**
+   * Resolve DTCG aliases when applying the input.
+   *
+   * @default true
+   */
   resolveAliases?: boolean;
+  /**
+   * Limit input application only to the listed sets.
+   *
+   * In combination with `modifiers`, this will limit output to only
+   * tokens declared within the options given. If tokens are referenced by alises
+   * outside these options, the application will fail unless `resolveAliases` is
+   * set to false.
+   */
+  sets?: string[];
+  /**
+   * Limit input application only to the listed modifiers.
+   *
+   * In combination with `sets`, this will limit output to only
+   * tokens declared within the options given. If tokens are referenced by alises
+   * outside these options, the application will fail unless `resolveAliases` is
+   * set to false.
+   */
+  modifiers?: string[];
 };
 
 export interface Resolver<
--- a/packages/parser/test/resolver.test.ts
+++ b/packages/parser/test/resolver.test.ts
@@ -1,9 +1,8 @@
 import fs from 'node:fs/promises';
 import * as momoa from '@humanwhocodes/momoa';
 import stripAnsi from 'strip-ansi';
-import { beforeEach, describe, expect, it } from 'vitest';
+import { describe, expect, it } from 'vitest';
 import defineConfig from '../src/config.js';
-import type { Resolver } from '../src/index.js';
 import Logger from '../src/logger.js';
 import parse from '../src/parse/index.js';
 import { calculatePermutations, validateResolver } from '../src/resolver/index.js';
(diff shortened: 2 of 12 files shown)


mrginglymus:
(obviously will add tests if the general principle is approved)

drwpow:
Yup I’m on board with this! I think it’s the user’s choice to want to only act on a subset of a resolver, the resolver design isn’t dependent on it being irreducible at every level. To me this is the same as simply deleting part of the document anyway, which should be the author’s choice to do so.

mrginglymus:
Added a couple of simple tests. I've added my go-to blue-orange/light-dark resolver for testing more intricate usage.

mrginglymus:
One limitation of this is that it still wont work if you have an `$extends` that references a token not in the current set of files. 

However, solving this would probably require something a bit more sophisticated implementation wise, whereby the root document is created from merging all sources, but recording in which source each token was declared. Then, at token resolution time, only _resolving_ tokens that come from sources on the allow list.

packages/parser/test/resolver.test.ts
```diff
@@ -458,6 +459,41 @@ describe('Resolver module', () => {
   });
 });
 
+describe('partial application', () => {
+  let resolver: Resolver;
```
drwpow: As a general rule we should try to not share state between tests. It’s a really hard mistake to undo when tests now have to execute in a certain order otherwise the entire suite breaks. Even if this is “safe” as-written, it only takes one modification or two here (likely by another author that doesn’t understand) the suite is poisoned with state contamination.

Instead, let’s move this to a pure `loadResolver()` helper that’s called within each test. Then each test sets up what it needs, and doesn’t affect other tests, and we still are DRY without sharing state.

mrginglymus: I just miss fixtures from pytest...will update :)

drwpow: Ah I didn’t realize it came from Python! Yeah I’ve just spent almost 10 years untangling leaky state from large test suites 🙃. Even in a smaller project like this library, it still turns into a real timesink when it breaks, unfortunately (especially because JS test runners like Vitest operate using a lot more magic than one might think)

drwpow:
> One limitation of this is that it still wont work if you have an `$extends` that references a token not in the current set of files.

That’s fine for now IMO. The two flavors of that—DTCG alias and JSON Schema `$ref`—both are expected to err if the resolution fails, and resolution is up to the user to manage. If it doesn’t resolve with partial inputs, then I think that’s desirable behavior for it to err as if that token never existed in the first place.

packages/parser/src/types.ts
```diff
@@ -351,6 +351,7 @@ export type ResolverInput = Record<string, string>;
 export type ResolverApplicationOptions = {
   sets?: string[];
   modifiers?: string[];
+  resolveAliases?: boolean;
```
drwpow: Can we add this to the documentation for the JS API? I want to make sure I’m understanding when I want this to be true or false. A short section & example would be great if you can provide one!

packages/parser/test/resolver.test.ts
mrginglymus: /wildly off topic

pytest has a full dependency injection framework, so you _never_ need to use an explicit `beforeEach` but instead you'd do something like:

```py

@fixture
def resolver():
  # parse your resolver here
  return resolver

@test
def test_resolver(resolver):
  # resolver fixture injected by name
  assert resolver
```

happy days. The fixtures can have different scopes to share resources (and save processing time), cleanups that will run even if tests fail. Oh, and it uses native python assertions with some _horrific_ byteweaving to add in rich failure messages.

packages/parser/src/types.ts
mrginglymus: @drwpow done - did some drive-by polishing of some adjacent dusty bits, but would welcome a redraft my redraft.

www/src/pages/docs/reference/js-api.md
```diff
@@ -270,25 +270,54 @@ import { createResolver, parse } from "@terrazzo/parser";
 const { resolver } = await parse(sources, { config });
 const r = createResolver(resolver);
 
-for (const input of r.inputPermutations) {
+for (const input of r.listPermutations() ?? [{}]) {
```
drwpow, line 273: Ah good catch, ty!

www/src/pages/docs/reference/js-api.md
```diff
@@ -270,25 +270,54 @@ import { createResolver, parse } from "@terrazzo/parser";
...
   r.apply(input); // tokens for this input
 }
 ```
 
-It’s up to you to specify all the permutations desired. `.getAllInputPermutations()` will return an array of all possible inputs for the specified modifiers.
```
drwpow, line 278: 🤦 Had quite a few old references to some early things that got renamed, I see. Thanks again

drwpow (approved):
Amazing, thank you for adding! Good with this merging, would you mind adding a `minor` changeset (see comment) so we can release it? Thanks!

drwpow (approved):
🎉

PR #764, opened 2026-06-01
fix: typography token aliasing another typography token crashes the listing build

unpunnyfuns:
A typography token whose `$value` aliases another typography token crashes the listing build: `TypeError: Cannot use 'in' operator to search for '.' in undefined`.

`transformTypography` hands `transformAlias` a bare `{ id }` stub for the alias target — fine for the default id-generator (reads only `id`), but consumers that recursively transform it (plugin-token-listing's preview computation) dereference its missing `mode` and throw. Fix: pass a complete sub-token (`id` + `$type` + `$value` + `mode`), so CSS output is unchanged and recursive consumers resolve a real value. The plugin also guards `rToken.mode` as defense-in-depth. End-to-end test included.


Sidnioulz (commented):
Good catch! Thanks for the PR @unpunnyfuns!

A question for @drwpow: how is the caller of `transformCSSValue` supposed to tell which mode it wants since TZ 2.0?

Here it seems the code swallows any mode option and returns the default mode, though the token listing code wants to transform a value for a specific mode. I suspect more changes are needed as a result.

packages/token-tools/src/css/typography.ts
```diff
@@ -74,3 +87,27 @@ export function transformTypography(
   }
   return output;
 }
+
+/** Map a typography sub-property to the `$type` its value transforms as. Mirrors the per-property switch above. */
+function typographySubValueType(property: string, subvalue: unknown): string {
```
drwpow: Minor suggestion: this is more type-safe (even if a little annoying to consume, it helps prevent typos)

```suggestion
function typographySubValueType(property: string, subvalue: unknown): Token['$type'] {
```

packages/plugin-token-listing/test/utils.test.ts
```diff
@@ -371,6 +371,35 @@ describe('computePreviewValue', () => {
     });
   });
 
+  describe('alias to a mode-less token', () => {
+    it('does not throw when the alias target lacks a mode map (typography aliasing typography)', async () => {
+      const { transformCSSValue } = await import('@terrazzo/token-tools/css');
+      // token-tools resolves a typography-to-typography alias by handing
+      // `transformAlias` a synthetic stub that has no `mode` map. The recursive
+      // transform must not evaluate `mode in rToken.mode` on that stub.
+      vi.mocked(transformCSSValue).mockImplementation((tok: any, opts: any) => {
```
drwpow: There’s no reason to mock this in tests. Especially with parser functions, this only gives false assurance something is working when it’s actually deviated from the runtime code. Let’s just call the functions themselves. We can test partial assertions if needed.

packages/plugin-token-listing/test/utils.test.ts
```diff
@@ -371,6 +371,35 @@ describe('computePreviewValue', () => {
...
+        if (tok.mode) {
+          // top-level call: recurse into the mode-less alias-target stub
```
drwpow: Nit: we don’t need comments everywhere—was this Claude-generated?

drwpow (commented):
Thanks! In favor of this PR merging, but I‘d like to mostly request we not mock anything in tests—there‘s no reason to mock anything that is self-contained. Mocking should only be done either when you’re testing an external system you have no control over, or you’re dealing with something that is both complex and ancillary to the thing tested. Neither scenario applies here, so let’s just execute actual runtime code for better tests.

unpunnyfuns:
Yes, afraid I unleashed the LLM on this one, as I caught it downstream. I should have taken greater care to review. Will not happen again.

drwpow:
> Good catch! Thanks for the PR @unpunnyfuns!
> 
> A question for @drwpow: how is the caller of `transformCSSValue` supposed to tell which mode it wants since TZ 2.0?
> 
> Here it seems the code swallows any mode option and returns the default mode, though the token listing code wants to transform a value for a specific mode. I suspect more changes are needed as a result.

Whoops sorry I missed this question! The idea of “modes” is soft-deprecated in 2.0, meaning it’s still supported but has limited application over resolver contexts. The way to access that is by using the [backwards-compatible syntax](https://terrazzo.app/docs/guides/resolver-contexts/#modes-legacy): `{ tzMode: 'myMode' }`. i.e. pretend there’s a modifier named `tzMode` (`tzMode` was chosen because simply “mode” is a common word that folks likely may have in their resolvers, and we didn’t want to have bugs from conflicts).

drwpow (approved):
Perfect! Thanks for the fixes! Appreciate you reworking the test to avoid mocking—that goes miles in helping us catch other bugs.

drwpow:
Just kicked off the release for this! It’ll be present in 2.3.0, ready in a few min

PR #765, opened 2026-06-02
fix(plugin-sass): forward CSS plugin's localID in transform hook

mrisoli:
## Summary

Fixes a regression in `@terrazzo/plugin-sass` where the CSS plugin's `variableName` customization (prefixes, casing) was not honored by the generated `$__token-values` Sass map.

plugin-sass's `transform` hook was setting `localID: token.id` (the dotted DTCG id), which `build.ts` then read via `getTransforms({ format: FORMAT_ID })` and turned into a CSS variable name from scratch — bypassing the CSS plugin's `variableName` and producing `var(--color-blue-40)` instead of `var(--ds-color-blue-40)`. The typography mixins map was unaffected because that loop queries `format: 'css'` directly.

Forwarding `token.localID` instead preserves any prefix or casing customization configured via `variableName` on plugin-css.

## Root cause

Regression introduced in 2.2.0 (PR #589) when the main loop in `build.ts` switched from `format: 'css'` to `format: FORMAT_ID`.

## Changes

- `packages/plugin-sass/src/index.ts`: forward `token.localID` instead of `token.id` in the `setTransform` call.
- Added changeset (`@terrazzo/plugin-sass`: patch).

### DIFF
--- a/.changeset/plugin-sass-forward-css-localid.md
+++ b/.changeset/plugin-sass-forward-css-localid.md
@@ -1,0 +1,9 @@
+---
+"@terrazzo/plugin-sass": patch
+---
+
+Forward the CSS plugin's `localID` when registering the sass transform, so any `variableName` customization (prefixes, casing) configured on plugin-css is honored by the generated `$__token-values` Sass map.
+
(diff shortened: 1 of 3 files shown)


drwpow (approved):
Great fix! Agree with just updating the existing test here—this should cover the behavior. Ashamed I didn’t notice the prefix was missing in the fixture!

PR #766, opened 2026-06-02
feat(plugin-token-listing): Allow multiple preview values

AngelRionCervi:
We have an issue with my team where we wanted to display a Compose Color and a SwiftUI Color token name and be able to also display the color in a CSS compatible format on a documentation page. Currently `previewValue` seems to only handle transformation to CSS.

Screenshot 2026-06-02 at 14 46 18

The idea is to replace the current `previewValue` with a `preview` object that stores different name formats for different stacks.

## Changes

Add the ability to generate multiple preview values in order to preview different token format. 
Also exports the current `computePreviewValue` renamed to `computeCssPreviewValue` so that it can be used alongside a custom function.

### DIFF
--- a/packages/plugin-token-listing/src/build.ts
+++ b/packages/plugin-token-listing/src/build.ts
@@ -1,7 +1,7 @@
 import type { Logger, ModeMap, Plugin, TokenNormalized, TransformParams } from '@terrazzo/parser';
 import type { TokenTransformed } from '@terrazzo/token-tools';
 import type { PlatformOption, TokenListing, TokenListingExtension, TokenListingPluginOptions } from './lib.js';
-import { computePreviewValue } from './utils/previewValue.js';
+import { computeCssPreviewValue } from './utils/preview.js';
 import mapValues from './utils/utils.js';
 
 export * from './lib.js';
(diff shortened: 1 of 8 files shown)


drwpow (approved):
Thanks! I’m open to this change, not sure if @Sidnioulz has feedback. Either way, could you add a changeset file (see comment) with a `patch` bump? (`patch` just because the token listing plugin is still 0.x, and we’re using minor versions for breaking changes)

Sidnioulz:
Hey there @AngelRionCervi!

If I understand correctly, you want to output a token listing intended for consumption on a non-CSS-compatible documentation platform?

The goal of `previewValue` is to show what a value looks like once rendered; not to show the canonical value as it is (e.g. hex8, Swift Color). For that, Token Listing 1.0 extends `platforms` to support a name _and a value_ per platform.

I have the code written up for Terrazzo but I'm still building it for Style Dictionary before releasing it. Would you be ok to use a patch TZ version for a couple more weeks until I finish the SD implem and am confident this is truly a 1.0?

AngelRionCervi:
Hi @Sidnioulz !

Yes the goal of this MR is basically to be able to add arbitrary values in the `preview` object so it can be previewed (visually and the value as string) on any chosen stacks.

We have a workaround at the moment with the team so it can wait the 1.0 if it brings the same per token custom preview functionality 🙂

Sidnioulz:
Ok, 1.0 should work for you then. `preview` answers the visual preview need (and assumes it is identical everywhere); whereas `platforms.*.value` answers the actual value showcase need.

I think we can close the PR @drwpow!

PR #767, opened 2026-06-03
Use shared token types for plugin-js declarations

heddendorp:
## Why

`@terrazzo/plugin-js` emits generated token package declarations that currently import normalized token types from `@terrazzo/parser`. That makes a generated token package expose the parser package as part of its consumer-facing type surface, even though consumers only need token data types.

The earlier inline approach fixed the parser dependency but introduced two real problems: it weakened `$value` typing and created a second copy of the normalized token model that could drift from Terrazzo's source of truth.

## How this fixes it

This adds a new dependency-free `@terrazzo/token-types` package for the shared public token type surface. The existing token definitions move there, with `@terrazzo/token-tools` re-exporting them so existing imports continue to work. Parser also keeps its convenience re-exports, now backed by the shared package.

`@terrazzo/plugin-js` generated `.d.ts` files now import `TokenNormalizedSet` and token-specific normalized aliases such as `ColorTokenNormalized` from `@terrazzo/token-types` instead of `@terrazzo/parser`. That preserves the current generated aliases and concrete `$value` types without requiring generated token package consumers to install or resolve parser internals.

The new type package has no runtime dependencies. Its `source.node` field is intentionally assignable without importing Momoa so parser internals can keep using AST nodes while the public token type package stays dependency-free.

## Verification

Added a regression test that builds a token package with `@terrazzo/plugin-js`, copies only the generated JS/declaration output plus `@terrazzo/token-types` into a clean consumer package, and runs TypeScript without installing `@terrazzo/parser`. The test also verifies a generated `Color` alias retains its normalized `$value.components` type.

Verified:

- `pnpm --filter @terrazzo/token-types lint`
- `pnpm --filter @terrazzo/token-types build`
- `pnpm --filter @terrazzo/token-tools lint`
- `pnpm --filter @terrazzo/token-tools build`
- `pnpm --filter @terrazzo/parser lint`
- `pnpm --filter @terrazzo/parser build`
- `pnpm --filter @terrazzo/plugin-js lint`
- `pnpm --filter @terrazzo/plugin-js test`
- `pnpm --filter @terrazzo/plugin-js build`


drwpow:
Thanks for opening! I don’t mind making this change in general, but I think this approach has two problems:

1. The types are incorrect—they’re not only missing parts of the shape from `@terrazzo/parser`, but they also don’t fill in the correct `value` types? I think we’d need significantly better types in order to not have a breaking change downstream
2. This now has the danger of drift from 2 competing sources of truth. We’re not running type tests on `plugin-js` (at least not many)

> That makes published token packages leak parser internals into consumers. A browser/Vite/TypeScript app importing generated token data then has to resolve @terrazzo/parser types, which can pull in parser-only dependencies such as `yaml-to-momoa` and Node globals such as `Buffer`.

The parser has no Node internals (it’s ESM and browser-compatible), and `yaml-to-momoa` is not installed via the parser—one has to provide one’s own. Vitest should be smart enough to separate type imports from runtime. But just in the interest of fixing this, what if we took an alternative approach: **made a new types package?**

If we had, say, `@terrazzo/token-types`, that was consumed by the parser and all plugins, would that satisfy the requirements? I don’t want to pursue “self-contained code” as a goal of this plugin, because that was never a design goal. But if we’re concerned about dependency chains, I think having a pure types package with no runtime, and no dependencies, should address the core concern without any of the drawbacks, I think?

heddendorp:
Thanks, that direction makes sense. I reworked the PR away from inline declarations and toward the shared package boundary you suggested.

What changed:

- added `@terrazzo/token-types` as a dependency-free package for the public token type surface
- moved the existing token type definitions there and made `@terrazzo/token-tools` re-export them, so existing imports keep working
- updated parser convenience re-exports to come from `@terrazzo/token-types`
- changed `@terrazzo/plugin-js` generated `.d.ts` files to import token types from `@terrazzo/token-types` instead of `@terrazzo/parser`
- updated the regression test so a generated token package type-checks in a clean consumer project with `@terrazzo/token-types`, but without `@terrazzo/parser`

This keeps the concrete normalized `$value` types and avoids creating an inline second source of truth in every generated declaration file. I also adjusted the PR description to call out why this is now a shared-types-package approach rather than a self-contained-output approach.

heddendorp:
The Node 24 test failure was `@terrazzo/plugin-sass:test` timing out while its CLI fixture tests ran `tz build`. I pushed a small follow-up that gives those CLI fixture tests an explicit `30_000ms` timeout, matching the existing pattern used for longer CLI fixture builds in `plugin-js`.

Additional local verification for that fix:

- `pnpm --filter @terrazzo/plugin-css build`
- `pnpm --filter @terrazzo/plugin-sass test`
- `pnpm --filter @terrazzo/plugin-sass lint`
- `pnpm --filter @terrazzo/plugin-sass build`

drwpow:
> The Node 24 test failure was `@terrazzo/plugin-sass:test` timing out while its CLI fixture tests ran `tz build`. I pushed a small follow-up that gives those CLI fixture tests an explicit `30_000ms` timeout

Perfectly reasonable, thank you! That runtime creep has nothing to do with this PR I’m sure and can be investigated separately.

drwpow (approved):
Great improvement, thank you!

PR #771, opened 2026-06-08
Fix Swift asset catalog component decimals

### LINKED ISSUE #768: Swift plugin emits alpha "1", which Xcode interprets as 1/255
## Description

`@terrazzo/plugin-swift` emits whole-number color components as strings such as `"alpha": "1"` in an asset catalog. Xcode interprets that value as the 8-bit integer `1`, producing alpha `1/255` instead of full opacity.

## Versions

- `@terrazzo/plugin-swift`: `0.3.1`
- `@terrazzo/cli`: `2.2.0`
- Xcode: `26.4.1`
- macOS deployment target: `26.3`

## Minimal token

```json
{
  "color": {
    "$type": "color",
    "surface": {
      "canvas": {
        "$value": {
          "colorSpace": "srgb",
          "components": [0.8588235, 0.8588235, 0.8588235],
          "alpha": 1
        }
      }
    }
  }
}
```

The same result occurs when `alpha` is omitted, because it defaults to `1`.

## Generated output

```json
"components": {
  "red": "0.8588235",
  "green": "0.8588235",
  "blue": "0.8588235",
  "alpha": "1"
}
```

After compiling the asset catalog, `assetutil --info Assets.car` reports:

```text
Color components: [0.8588234782, 0.8588234782, 0.8588234782, 0.0039215686]
```

The alpha is therefore `1/255` rather than `1.0`.

## Expected output

The plugin should emit decimal strings for normalized components, for example:

```json
"alpha": "1.0"
```

Xcode then compiles the color with alpha `1.0` as expected. The same normalization should apply to exact zero/one RGB components (`"0.0"` and `"1.0"`) to avoid integer interpretation.

marko1olo:
## Changes

Closes #768.

- Format exact  and 1 Swift asset-catalog color components as "0.0" and "1.0" so Xcode treats them as normalized component values instead of 8-bit integer values.
- Keep non-integer component precision unchanged.
- Add an explicit regression assertion that generated asset catalog component strings do not contain bare "0" or "1" values.
- Update plugin-swift snapshots and add a patch changeset.

## How to Review

Useful review points:

- ormatAssetCatalogComponent() only changes exact normalized endpoints, so existing fractional values stay byte-for-byte as String(value) output.
- Snapshot diffs should only show "0" -> "0.0" and "1" -> "1.0" for 
ed/green/lue/lpha components.

Checks run:

- pnpm run build from repo root passed after adding a temporary local pnpm.cmd shim because Turbo cannot find a package-manager binary when pnpm is invoked as corepack pnpm on this Windows checkout.
- pnpm --filter @terrazzo/plugin-swift run test passed.
- pnpm --filter @terrazzo/plugin-swift run build passed.
- pnpm exec biome check packages/plugin-swift/src/index.ts packages/plugin-swift/test/index.test.ts packages/plugin-swift/test/color/Want.xcassets/color.primitive.blue.1.colorset/Contents.json .changeset/swift-alpha-decimal-components.md passed.
- git diff --check HEAD~1 HEAD passed.

Note: full pnpm --filter @terrazzo/plugin-swift run lint on this Windows worktree fails on CRLF formatting in untouched package files after checkout/restore. I did not include those EOL-only changes in this PR.


packages/plugin-swift/src/index.ts
```diff
@@ -27,6 +27,13 @@ export interface ColorData {
   };
 }
 
+function formatAssetCatalogComponent(value: number): string {
+  if (value === 0 || value === 1) {
```
drwpow, line 31: To be safer, let’s do the following instead: we’ll use JavaScript’s built-in formatter rather than manually creating our own string, and we’ll handle any integer, not just 0 or 1

```suggestion
  if (value % 0 === 0) {
    return value.toFixed(1);
```

packages/plugin-swift/test/index.test.ts
```diff
@@ -28,6 +28,11 @@ describe('@terrazzo/plugin-swift', () => {
     const result = await build(tokens, { config, resolver, sources });
     for (const { filename, contents } of result.outputFiles) {
       await expect(contents).toMatchFileSnapshot(fileURLToPath(new URL(filename, cwd)));
+      const { colors } = JSON.parse(contents) as { colors?: { color?: { components?: Record<string, string> } }[] };
+      for (const color of colors ?? []) {
+        expect(Object.values(color.color?.components ?? {})).not.toContain('0');
+        expect(Object.values(color.color?.components ?? {})).not.toContain('1');
```
drwpow, line 34: 👍

drwpow (commented):
Thanks for the fix! I’d like to keep this but make one small adjustment to make this a little more resilient (see comment). With that change, happy to merge & release! 🙂

drwpow (approved):
Will make the requested changes manually. Thanks again!

PR #772, opened 2026-06-08
Preserve zero dimension units

### LINKED ISSUE #747: `plugin-vanilla-extract` strips the unit off zero-valued dimension tokens
### Versions

- @terrazzo/cli: 2.0.3
- @terrazzo/plugin-css: 2.0.3
- @terrazzo/plugin-vanilla-extract: 2.0.3
- @terrazzo/token-tools: 2.0.3

### Node.js version

24.13.0

### OS + version

macOS 26.4.1

### Description

Hi! I'm hitting an issue where zero-valued dimension tokens lose their unit in the generated `theme.css.ts`. A token I'm authoring as `{ "$type": "dimension", "$value": { "value": 0, "unit": "px" } }` ends up as `'0'` in the output rather than `'0px'`.

I think this is happening in `serializeValue`. There's a regex that matches `"0<unit>"` strings and rewrites them to just `"0"`. The same behavior looks like it's in `@terrazzo/token-tools`' `transformDimension` too, so folks using `plugin-css` on its own would hit it.

I know `margin: 0` is valid CSS, so bare `0` works fine when the token gets used directly in a property. Where it breaks for me is when a downstream consumer composes the token into a `calc()`. Per [CSS Values L4 § Math Functions](https://www.w3.org/TR/css-values-4/#calc-type-checking), something like `calc(21px + 0)` isn't valid and the browser drops the declaration. In my case, VE's theme values are going out as CSS custom properties, so consumer code that does `` `calc(20px + ${vars.space.none})` `` is silently broken at runtime, with no build error or console warning.

I tried working around it on my side with a custom `transform` on `plugin-css` that returns `'0px'`, but `serializeValue` strips the unit back off before it writes `theme.css.ts`, so I couldn't find a way out user-side.

### Reproduction

`tokens.json`:

```json
{
  "space": {
    "none":   { "$type": "dimension", "$value": { "value": 0, "unit": "px" } },
    "medium": { "$type": "dimension", "$value": { "value": 8, "unit": "px" } }
  }
}
```

`terrazzo.config.mjs`:

```js
import css from '@terrazzo/plugin-css';
import ve from '@terrazzo/plugin-vanilla-extract';

export default {
  tokens: ['./tokens.json'],
  outDir: './out',
  plugins: [
    css({ filename: 'tokens.css' }),
    ve({
      filename: 'theme.css.ts',
      globalThemes: { base: { selector: ':root', mode: '.' } },
    }),
  ],
};
```

Running `npx tz build`, I get this in `out/theme.css.ts`:

```ts
export const base = createGlobalTheme(':root', vars, {
  space: {
    none: '0',     // would expect '0px' here
    medium: '8px',
  },
});
```

And here's the consumer pattern that breaks silently:

```ts
import { vars } from './out/theme.css';
import { style } from '@vanilla-extract/css';

export const row = style({
  // Renders as calc(20px + var(--space-none) + var(--space-medium))
  // which resolves at runtime to calc(20px + 0 + 8px) and gets discarded.
  height: `calc(20px + ${vars.space.none} + ${vars.space.medium})`,
});
```

### Expected result

Zero-valued dimension tokens keep their unit, so they're safe to drop into `calc()` on the consumer side:

```ts
space: { none: '0px', medium: '8px' }
```

I'd be happy to open a PR if it'd help! I don't have a strong preference between just removing the zero-strip or putting it behind an option (maybe something like `minifyZeroLengths: false`). Whatever shape you'd prefer, just let me know.

Thanks!

### Extra

- [x] I'm willing to open a PR (see [CONTRIBUTING.md](https://github.com/terrazzoapp/terrazzo/blob/main/CONTRIBUTING.md))

marko1olo:
## Changes

Closes #747.

- Preserve the original unit for zero-valued DTCG dimension tokens in `@terrazzo/token-tools` instead of serializing them as bare `0`.
- Remove the duplicate zero-unit stripping in `@terrazzo/plugin-vanilla-extract`.
- Update plugin-css and vanilla-extract snapshots so generated custom-property values like `0px` / `0rem` remain type-compatible when consumers compose them in CSS math functions such as `calc()`.
- Add a patch changeset for `@terrazzo/token-tools`, `@terrazzo/plugin-css`, and `@terrazzo/plugin-vanilla-extract`.

## How to Review

The main behavior change is in `packages/token-tools/src/css/dimension.ts`. The snapshot churn is expected: dimension-derived zero values now keep their unit in CSS and vanilla-extract output.

Validated locally:

- `corepack pnpm --filter @terrazzo/token-tools run test`
- `corepack pnpm --filter @terrazzo/plugin-css exec vitest run --project unit`
- `corepack pnpm --filter @terrazzo/plugin-vanilla-extract run test`
- `corepack pnpm --filter @terrazzo/token-tools run lint:ts`
- `corepack pnpm --filter @terrazzo/plugin-css run lint:ts`
- `corepack pnpm --filter @terrazzo/plugin-vanilla-extract run lint:ts`
- `corepack pnpm exec biome check --formatter-enabled=false --diagnostic-level=error packages/token-tools/src/css/dimension.ts packages/token-tools/test/css.test.ts packages/plugin-vanilla-extract/src/lib.ts packages/plugin-vanilla-extract/test/lib.test.ts .changeset/zero-dimension-units.md`
- `corepack pnpm --filter @terrazzo/token-tools run build`
- `corepack pnpm --filter @terrazzo/plugin-css run build`
- `corepack pnpm --filter @terrazzo/plugin-vanilla-extract run build`

Also attempted `corepack pnpm --filter @terrazzo/plugin-css exec vitest run --project browser` after installing Chromium; on this Windows workstation Vitest failed before loading tests with `ERR_CONNECTION_REFUSED` to its local browser runner URL.


drwpow (approved):
Perfect! No notes, all snapshot updates seem reasonable and correct. The original issue explains full context and rationale behind this change.

PR #773, opened 2026-06-08
Fix setTransform for modifier-only tokens

marko1olo:
## Summary

Fixes #707.

`setTransform()` currently validates and stores transformed-token metadata from the base token set passed into `build()`. With a resolver, that base set is the default permutation, so a plugin cannot transform a token that only exists in a non-default modifier context even when it supplies `input` for that context.

This updates `setTransform()` to resolve token metadata from `resolver.apply(params.input)` for non-legacy resolver inputs, while preserving the existing base-token behavior for default and legacy `tzMode` transforms.

## Tests

- `corepack pnpm --filter @terrazzo/parser exec vitest run test/plugin.test.ts -t "resolver input"`
- `corepack pnpm --filter @terrazzo/parser run test`
- `corepack pnpm --filter @terrazzo/parser run lint:ts`
- `corepack pnpm exec biome check --formatter-enabled=false --diagnostic-level=error packages/parser`
- `corepack pnpm --filter @terrazzo/parser run build`
- `git diff --check`

Assisted-by: OpenAI Codex

### DIFF
--- a/.changeset/parser-resolver-input-transforms.md
+++ b/.changeset/parser-resolver-input-transforms.md
@@ -1,0 +1,5 @@
+---
+"@terrazzo/parser": patch
+---
+
+Fix `setTransform()` for resolver tokens that only exist in a non-default modifier input.
--- a/packages/parser/src/build/index.ts
+++ b/packages/parser/src/build/index.ts
@@ -122,14 +122,15 @@ export default async function build(
             });
             return;
           }
-          const token = tokens[id]!;
-          if (!token) {
-            logger.error({ group: 'plugin', label: plugin.name, message: `No token "${id}"` });
-          }
           const isLegacyModes = params.input && Object.keys(params.input).length === 1 && 'tzMode' in params.input;
           const permutationID =
             params.input && !isLegacyModes ? resolver.getPermutationID(params.input) : FALLBACK_PERMUTATION_ID;
           const mode = params.mode || (isLegacyModes && params.input.tzMode) || undefined;
+          const tokenSource = params.input && !isLegacyModes ? resolver.apply(params.input) : tokens;
+          const token = tokenSource[id]!;
+          if (!token) {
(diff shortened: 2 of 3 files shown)


mrginglymus:
Duplicate of #708?

marko1olo:
There is overlap with #708, yes. I opened this as a narrower version of the same fix rather than an independent feature.

The difference is that this PR only switches the token lookup source when `params.input` is a non-legacy resolver input. It keeps the existing fallback token set for no-input transforms and for the legacy `{ tzMode }` path, so it avoids the CSS fixture changes that #708 currently produces. The regression also asserts the `transform.token` object for a token that exists only in `{ theme: "dark" }`, not just the transformed value lookup.

If you prefer #708 as the base, the useful part from this PR is probably just that `tokenSource = params.input && !isLegacyModes ? resolver.apply(params.input) : tokens` distinction plus the modifier-only token regression.

drwpow:
Closing as duplicate of #708. Pretty sure @marko1olo is a bot (or a human account that uses a bot to automate opening PRs on GitHub repos at a high rate). @marko1olo before opening future PRs, please make sure there’s not an existing PR open addressing the issue!

marko1olo:
Hi @drwpow,

Fair enough to call that out — I owe you a straight answer.

Yes, I use AI coding tools heavily (GitHub Copilot, Claude, Gemini). But I am a real person, not a bot account. I read the source, I trace the actual bug, I verify the fix makes sense before opening a PR. That said, I have clearly been moving too fast, opening too many PRs in a short window, and not being careful enough about duplicate checking or the signal-to-noise ratio I'm adding to your repo.

The comments on #805 that you called "AI slop" — you're right, those are low-value filler that I should have stripped. I'll clean those up now.

I'll slow down, be more selective, and make sure each PR is genuinely worth your review time before I open it. Thanks for being direct about it.

Issue #759, opened 2026-05-16
Parser: support for orthogonality


drwpow:
### Summary

This is the more explicit ask of #654. Without knowing whether or not modifiers are _orthogonal_, we can’t deduplicate tokens (because tracing the graph is too complex in some cases it’s not practical). However, if the parser marks an entire resolver as `orthogonal: true`, we can allow certain capabilities downstream.

## Orthogonality

An entire resolver is considered orthogonal if all `modifiers` act on unique tokens. 

For example, if there’s a `theme` modifier that acts on the `color.text` token for light/dark mode, and there’s a `high-contrast` modifier that overrides `color.text` for high contrast mode only when `true`, that entire resolver is **NOT** orthogonal. It is not possible to say “a resolver is orthogonal in some permutations, but not in others”—that is by definition **non-orthogonal.** A resolver’s modifiers all either act on unique tokens, always, or they don’t.

Further, orthogonality is **a property of the resolver.** Just describing 2 modifiers as orthogonal, when a 3rd isn’t, means the entire resolver is **non-orthogonal.**

Lastly, orthogonality keys off ID, not value. If 2 modifiers touch the same IDs, and produce the same values, they are still **non-orthogonal.** Though, this is a unique situation that probably needs another name, because in some practices this does allow deduplication. But this is a more complex edge case, for now, we’re not considering relevant.

## Approach

Step 1 is adding an `orthogonal: boolean` property to `@terrazzo/parser` that attaches on all resolvers.

To achieve this, the parser must scan all IDs present in a modifier, in all contexts. Then we simply compare overlap between the modifiers. In this way **we can add support for orthogonality without doing any extra work.** We wouldn’t increase runtime, and we could still support orthogonality even if a resolver has 10k permutations. _Note: this is also why we’re saying orthogonality is based on ID, not value—if it were based on value, this would be orders of magnitude more complicated._

Step 2 is resolving #654, adding the ability for the CSS plugin to output selectors based on individual modifiers, rather than permutations. If the resolver is `orthogonal: false`, an error MUST be thrown as it could not generate usable code. But if it’s `orthogonal: true`, a user could opt into this deduplicated mode.

### Proposal & prior art

https://www.designtokens.org/tr/2025.10/resolver/#orthogonality

### Extra

- [ ] I’m willing to open a PR (see [CONTRIBUTING.md](https://github.com/terrazzoapp/terrazzo/blob/main/CONTRIBUTING.md))

[...]

PR #774, opened 2026-06-08
Add resolver orthogonality metadata

marko1olo:
## Summary

Implements step 1 of #759.

This adds `resolver.orthogonal: boolean` to `@terrazzo/parser`. The value is computed once when the resolver is created by scanning token IDs touched by each modifier across all of its contexts:

- modifiers that touch disjoint token IDs -> `orthogonal: true`
- any token ID touched by more than one modifier -> `orthogonal: false`
- repeated IDs within the same modifier are allowed, since contexts of one modifier are mutually exclusive

This deliberately does not change CSS output yet. It exposes the parser metadata needed by downstream plugins to safely opt into modifier-scoped output later.

## Tests

- `corepack pnpm --filter @terrazzo/parser exec vitest run test/resolver.test.ts -t "orthogonal"`
- `corepack pnpm --filter @terrazzo/parser run test`
- `corepack pnpm --filter @terrazzo/parser run lint:ts`
- `corepack pnpm exec biome check --formatter-enabled=false --diagnostic-level=error packages/parser`
- `corepack pnpm --filter @terrazzo/parser run build`
- `git diff --check`

Assisted-by: OpenAI Codex

### DIFF
--- a/packages/parser/src/resolver/load.ts
+++ b/packages/parser/src/resolver/load.ts
@@ -1,12 +1,19 @@
 import type * as momoa from '@humanwhocodes/momoa';
-import { type InputSource, type InputSourceWithDocument, maybeRawJSON } from '@terrazzo/json-schema-tools';
+import { type InputSource, type InputSourceWithDocument, maybeRawJSON, parseRef } from '@terrazzo/json-schema-tools';
 import type { TokenNormalizedSet } from '@terrazzo/token-tools';
 import type yamlToMomoa from 'yaml-to-momoa';
 import { toMomoa } from '../lib/momoa.js';
 import { destructiveMerge, getPermutationID } from '../lib/resolver-utils.js';
 import type Logger from '../logger.js';
 import { processTokens } from '../parse/process.js';
-import type { ConfigInit, Resolver, ResolverInput, ResolverSourceNormalized } from '../types.js';
+import { aliasToGroupRef } from '../parse/token.js';
+import type {
+  ConfigInit,
+  Resolver,
+  ResolverInput,
+  ResolverModifierNormalized,
+  ResolverSourceNormalized,
+} from '../types.js';
 import { normalizeResolver } from './normalize.js';
 import { isLikelyResolver, validateResolver } from './validate.js';
 
@@ -106,6 +113,7 @@ export function createResolver(
   const inputDefaults: ResolverInput = {};
   const validContexts: Record<string, string[]> = {};
   const allPermutations: ResolverInput[] = [];
+  const modifiers: ResolverModifierNormalized[] = [];
 
   const resolverCache: Record<string, any> = {};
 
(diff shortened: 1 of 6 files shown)


packages/parser/src/resolver/load.ts
```diff
@@ -232,6 +247,45 @@ export function createResolver(
...
+  const tokenIDs = new Set<string>();
+  for (const sources of Object.values(modifier.contexts)) {
+    for (const source of sources) {
+      collectTokenIDs(source, [], tokenIDs);
+    }
+  }
+  return tokenIDs;
+}
+
+function collectTokenIDs(source: unknown, path: string[], tokenIDs: Set<string>): void {
```
mrginglymus: This will need some way to handle `$extends`. In my implementations I pass in the flattened default token structure and copy out extends from there. It does rely somewhat on each `$extends` usage generating the exact same structure under any input which is not guaranteed.

marko1olo: Thanks, agreed. I updated this in `1e668664` so the orthogonality scan now builds the default raw token structure, overlays each modifier context before collecting IDs, and follows `$extends` targets at the current path.

I also added a regression where two modifiers both touch `component.button` through `{component.defaults}`; `resolver.orthogonal` is now `false` for that case.

Rechecked locally:
- `node_modules\.bin\vitest.cmd run packages\parser\test\resolver.test.ts packages\parser\test\extends.test.ts`
- `node_modules\.bin\tsc.cmd --noEmit -p packages\parser\tsconfig.json`
- `node_modules\.bin\biome.cmd check packages\parser\src\resolver\load.ts packages\parser\test\resolver.test.ts`
- `git diff --check`

marko1olo:
Pushed `262a3c10` to fix the CI failure in `@terrazzo/plugin-css#test`.

The failing fixture merged a typography mode where the default token had a primitive `lineHeight: 1.5`, while another mode replaced that field with an object. The orthogonality scan builds a default token source and merges modifier contexts into it, so `destructiveMerge()` needed to treat object-over-primitive the same way it already treats array overwrite: replace the existing value before recursing.

Local checks:

- `vitest run test/lib/resolver-utils.test.ts test/resolver.test.ts` from `packages/parser` -> 35 passed
- `biome check packages/parser/src/lib/resolver-utils.ts packages/parser/test/lib/resolver-utils.test.ts packages/parser/src/resolver/load.ts packages/parser/test/resolver.test.ts` -> passed
- `tsc --noEmit -p tsconfig.json` from `packages/parser` -> passed
- `pnpm -r build` -> passed
- `node ../../../../cli/bin/cli.js build` from `packages/plugin-css/test/fixtures/mode-type-typography` -> passed, generated CSS matches the fixture snapshot with normalized line endings
- `git diff --check origin/main..HEAD` -> passed

I also checked local process cleanup after the Vitest timeout; no `terrazzo-759` test/build processes were left running.

packages/parser/src/resolver/load.ts
```diff
@@ -117,12 +125,19 @@ export function createResolver(
         inputDefaults[m.name] = m.default!;
       }
       validContexts[m.name] = Object.keys(m.contexts);
+      modifiers.push(m);
     }
   }
 
   const permutationCount = Object.values(validContexts).reduce((acc, context) => acc * context.length, 1);
+  const defaultTokenSource = getDefaultTokenSource(resolverSource, inputDefaults);
+  const modifierTokenIDs = new Map(
```
mrginglymus, line 134: this modifier information (list of all tokens declared within a modifier) is extremely valuable information to downstream plugins, and will be necessary to achieve the onward goals of this PR (emitting a per-context-per-modifier class with only those tokens), so storing it on the `Resolver` would be a big benefit if we're already computing it here.

drwpow (commented):
This is a good attempt at orthogonality parsing, but this creates a performance concern of doing a 2nd pass over all the source files again, which is going to be slow. Instead, this should happen during the same initial pass we collect and assemble tokens in. In `packages/parser/src/parse/process.ts#processTokens`, for example, we could modify the return signature to have token IDs split by set and modifier, which it could do because it knows it’s a resolver with `isResolver: true`.

This PR won’t be accepted unless it reuses the existing token parsing step(s) to assemble these IDs.

drwpow (changes requested):
See comment on performance concern

drwpow:
Closing in favor of #778

PR #775, opened 2026-06-08
Update DTCG spec links in token types docs

mrginglymus:
The old links a. pointed at the old github url which b. pointed to the draft spec, not the latest published version and c. hadd an off-by-one error in the composite types.


drwpow (approved):
Thank you!

Issue #781, opened 2026-06-16
@terrazzo/token-tools wonky types


mrginglymus:
### Versions

- @terrazzo/token-tools@2.4.0

### Description

[`types-cmznBsLU.d.ts`](https://npmx.dev/package-code/@terrazzo/token-tools/v/2.4.0/dist%2Ftypes-cmznBsLU.d.ts) contains the following content:

```ts
import * as import__terrazzo_token_types from "@terrazzo/token-types";
//#endregion
export { __exportAll as n, __reExport as r, types_d_exports as t };
//# sourceMappingURL=types-cmznBsLU.d.ts.map
```

which is highly wonky

### Extra

- [x] I’m willing to open a PR (see [CONTRIBUTING.md](https://github.com/terrazzoapp/terrazzo/blob/main/CONTRIBUTING.md))

mrginglymus:
`rolldown-plugin-dts` strikes again :(

mrginglymus:
Given no external bundling is actually done: 
https://github.com/terrazzoapp/terrazzo/blob/b032980a503033c5dafcf60464470327f65add36/rolldown.ts#L12-L16
have you considered just compiling with `tsc`?

mrginglymus:
I've had this feeling of _déjà vu_ before: https://github.com/terrazzoapp/terrazzo/pull/595#discussion_r2572617497

PR #782, opened 2026-06-16
poc use tsc

mrginglymus:
## Changes

Addresses #781 with a drive-by adding of missing `.npmignore`, and a switch to `tsc` for building instead of `rolldown`.

### DIFF
--- a/tsconfig.build.json
+++ b/tsconfig.build.json
@@ -1,0 +1,9 @@
+{
+  "extends": "./tsconfig.base.json",
+  "compilerOptions": {
+    "outDir": "${configDir}/dist",
+    "rootDir": "${configDir}/src"
+  },
+  "include":["${configDir}/src"],
+  "exclude": ["**/dist/*", "**/node_modules/*"]
+}
--- a/.changeset/true-masks-ring.md
+++ b/.changeset/true-masks-ring.md
@@ -1,0 +1,5 @@
+---
+"@terrazzo/token-types": patch
+---
+
+Don't ship src files for @terrazzo/token-types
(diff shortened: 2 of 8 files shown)


tsconfig.build.json
```diff
@@ -0,0 +1,9 @@
+{
+  "extends": "./tsconfig.base.json",
+  "compilerOptions": {
+    "outDir": "${configDir}/dist",
+    "rootDir": "${configDir}/src"
+  },
+  "include":["${configDir}/src"],
```
drwpow, line 7: Whoa TIL about `${configDir}`. Cool!

drwpow (approved):
Thanks for testing this out! I know we’ve talked about switching to TypeScript, I’d be on board but would want to make sure we’re using the Go version if we switch to it everywhere (which is 7? I haven’t kept up with the details of how to opt in to tsgo).

My main gripe with TypeScript is as soon as you are dealing with any sort of non-JS file (such as `.css` in React), you’ve ejected and gone to Rolldown. Even any sort of wasm interop can be a PITA. So TypeScript is sort of a long-term commitment that this package will only be pure JS, and nothing else, and sometimes I don’t want to make that commitment. However, given the stable state of Terrazzo 2.0, I’m actually happy with a lot of package responsibilities so I’d feel OK about running everything on pure TypeScript because that’s how it’s shaken out. But again, any way we can opt into tsgo as soon as possible, the better! And that can happen in followups.

tsconfig.build.json
mrginglymus: Never mind rewriting in go, this is the best typescript feature in recent years ;)

PR #791, opened 2026-07-06
Update to typescript 7; use for all non-bundled packages

mrginglymus:
## Changes

Switch to latest TS to get go version

## How to Review

It builds, I guess. There's a windows bug in the current version of turbo used so I'm going to do another PR updating that (and fixing any other windows-related build issues) before I start switching other packages to use tsc.

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,8 @@
     "stylelint-config-standard": "^40.0.0",
     "stylelint-order": "^8.1.1",
     "turbo": "^2.10.4",
-    "typescript": "^6.0.3",
+    "typescript": "npm:@typescript/typescript6@^6.0.0",
+    "typescript-7": "npm:typescript@^7.0.2",
     "vite": "catalog:",
     "vitest": "catalog:"
   },
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -36,8 +36,8 @@
     "terrazzo": "bin/cli.js"
   },
   "scripts": {
-    "build": "rolldown -c && attw --profile esm-only --pack . && vite build",
-    "dev": "rolldown -w -c",
+    "build": "tsc -b tsconfig.build.json && attw --profile esm-only --pack .",
+    "dev": "tsc -b tsconfig.build.json -w",
     "fmt": "oxfmt",
     "fmt:check": "oxfmt --check",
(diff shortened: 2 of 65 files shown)


mrginglymus:
@drwpow how would you feel about adding `DOM` to the root `tsconfig` `lib` option, and `node` to the root `types` field?

drwpow:
> @drwpow how would you feel about adding `DOM` to the root `tsconfig` `lib` option, and `node` to the root `types` field?

DOM I think is fine, `node` may not be needed? All core plugins + APIs are designed to work in a pure browser environment, or edge worker (like Cloudflare) where Node.js isn’t available. It’s really only the CLI, Tailwind, and Sass where there’s a Node.js env assumed. So it’s really the minority of packages that should need Node types, I think?

mrginglymus:
Grand, I'll double check where I encountered them. It may be that test files needed node, but production code doesn't. 

I'll rebase this tomorrow including the just-today released typescript 7.0.2

drwpow:
> Grand, I'll double check where I encountered them. It may be that test files needed node, but production code doesn't.
> 
> I'll rebase this tomorrow including the just-today released typescript 7.0.2

Ohhh that’s right—I think by accident we were basically ignoring test files from the TS config, which isn’t great.

We have tests to ensure that everything still runs in browsers, so I don’t really care if the root tsconfig gets Node types if that makes tests easier! Not a huge deal either way.

Also sorry for the noise! I had been wanting to switch to Oxlint and found time to this week 😅. Figured just having everything in the Vite ecosystem is easier

mrginglymus:
@drwpow all done!

All packages that don't do something with css now use typescript for building. All configurations use `ESNext` and `DOM` libs; all test configurations use `node` types, all production configurations do not by default but there are some exceptions which should be easily spottable from their `tsconfig.build.json` files.

drwpow:
> @drwpow all done!
> 
> All packages that don't do something with css now use typescript for building. All configurations use `ESNext` and `DOM` libs; all test configurations use `node` types, all production configurations do not by default but there are some exceptions which should be easily spottable from their `tsconfig.build.json` files.

Great! Sorry for the delay in reviewing, happy to merge & ship! Thanks so much 🙏

drwpow (approved):

PR #803, opened 2026-07-29
Escape */ in token descriptions to keep generated CSS valid

### LINKED ISSUE #802: A token $description containing */ closes the generated CSS comment early and breaks the stylesheet
### Describe the bug

`plugin-css` writes a token's `$description` into a `/* */` comment without escaping it, so a description containing `*/` closes the comment early and everything after it is parsed as CSS.

### Reproduction

`tokens.json`:

```json
{
  "color": {
    "$type": "color",
    "blue": {
      "$value": { "colorSpace": "srgb", "components": [0, 0.2, 1] },
      "$description": "Use for links */ } body { display: none;"
    },
    "primary": {
      "$value": "{color.blue}"
    }
  }
}
```

The emitted stylesheet closes the `:root` block at the `*/` inside the comment. Every custom property declared after that token lands in whatever selector the description text happens to form — in this example a `body { display: none; }` rule — and the trailing `*/` is left over as a parse error.

### Expected behaviour

The description is a free-text field, so any value in it should be emitted safely or omitted. A token description should not be able to change which selector the following variables belong to.

### Cause

`printNode` in `packages/plugin-css/src/lib.ts` interpolates the comment directly:

```ts
if (node.comment) {
  output += `${indent}/* ${node.comment} */\n`;
}
```

### Notes

This is reachable by anyone who writes a description containing `*/`, which is easy to do accidentally in prose about CSS — for example a description that mentions a code fragment. It does not require a hostile author, but it does mean a token file from an untrusted source can decide the contents of the generated stylesheet, so it is worth treating as more than a formatting nit.

I have a patch with a regression test and a changeset, and can open a PR against this issue.

marko1olo:
Closes #802.

### What

`printNode` interpolated `node.comment` straight into a `/* */` block, so a `$description` containing `*/` terminated the comment early and the rest of the description was parsed as CSS — closing `:root` and moving every subsequent custom property into whatever selector the text formed.

The fix replaces the sequence with `*\/`, which is inert inside a comment and keeps the description readable rather than dropping or truncating it.

### Tests

One case added to `packages/plugin-css/test/lib.test.ts`, asserting the emitted comment no longer contains a closing sequence and that the following variable is still inside `:root`. The plugin-css package is 54 passing.

### Note

I checked whether the same pattern exists in the sibling emitters before scoping this to `plugin-css`, and left them alone: only this one writes a free-text description into a block comment. If you would rather the escaping lived in a shared helper so `plugin-sass` and friends inherit it if they ever grow the same feature, say so and I will move it.

A changeset is included.

### DIFF
--- a/packages/plugin-css/src/lib.ts
+++ b/packages/plugin-css/src/lib.ts
@@ -194,7 +194,7 @@ export function printNode(
 
   if (node.type === 'Declaration') {
     if (node.comment) {
-      output += `${indent}/* ${node.comment} */\n`;
+      output += `${indent}/* ${node.comment.replaceAll('*/', String.raw`*\/`)} */\n`;
     }
     output += `${indent}${node.property}: ${node.value};\n`;
     return output;
--- a/packages/plugin-css/src/lib.ts
+++ b/packages/plugin-css/src/lib.ts
@@ -188,7 +188,8 @@ export function printNode(
 
   if (node.type === 'Declaration') {
     if (node.comment) {
-      output += `${indent}/* ${node.comment} */\n`;
+      // break any `*/` in the comment, otherwise it closes the comment early and the rest is parsed as CSS
+      output += `${indent}/* ${node.comment.replaceAll('*/', String.raw`*\/`)} */\n`;
     }
     output += `${indent}${node.property}: ${node.value};\n`;
     return output;
--- a/.changeset/css-description-comment-escape.md
+++ b/.changeset/css-description-comment-escape.md
@@ -1,0 +1,5 @@
+---
+"@terrazzo/plugin-css": patch
+---
+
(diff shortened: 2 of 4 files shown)


packages/plugin-css/src/lib.ts
```diff
@@ -188,7 +188,8 @@ export function printNode(
 
   if (node.type === 'Declaration') {
     if (node.comment) {
-      output += `${indent}/* ${node.comment} */\n`;
+      // break any `*/` in the comment, otherwise it closes the comment early and the rest is parsed as CSS
```
drwpow: Please remove the comment, it’s not needed here.

drwpow (commented):
This is a good fix, thank you! One minor request to delete the code comment, otherwise looks good 🙂

marko1olo:
Cleaned up and removed the comments — pushed cleanly!

Issue #804, opened 2026-07-29
Five core lint rules accept an option or declare a guard that does not take effect


marko1olo:
### Description

Five defects in the built-in `core/*` lint rules, all of the same kind: an option or a guard that does not do what it is declared to do. I found them together and they are all small, so I have listed them in one issue rather than five — say the word if you would rather split them.

Each was checked by running the rule through `parse()` with a config, and I have noted what the repo's own fixtures currently do, because in two cases that is the more telling part.

---

**1. `core/duplicate-values` reports aliases as duplicates of the token they alias.**

The guard is `if (typeof t.aliasOf === 'string' && isAlias(t.aliasOf)) continue`. `aliasOf` holds a resolved token ID — `parse/token.ts:290` assigns `refToTokenID(...)`, which returns `color.blue.100` — while `isAlias()` matches the `{color.blue.100}` reference form. So the second half is never true and every alias is compared against its target. The rule's own description says "excludes aliases".

The check is also in the wrong place: only the primitive branch consults it, while colors, dimensions, shadows and the rest take the structural-comparison branch below, which has no alias check at all. Since tokens are visited in sorted-ID order the report can name the original rather than the alias.

**This one your own fixture already catches.** `[core/duplicate-values] no duplicates` (`test/lint.test.ts:242`) is exactly the alias case — `color.blue.300` is `{color.blue.100}` — and it emits `color.blue.300 declared a duplicate value` today. It passes anyway, see item 5.

**2. `core/duplicate-values` never detects duplicate `duration` values.**

`duration` sits in the primitive list and is therefore compared with `values[t.$type].has(t.$value)`. A normalized duration value is the object `{ value, unit }`, so `Set.has` compares object identity and never matches — two tokens with `{ value: 100, unit: 'ms' }` are not reported. `dimension` is already excluded from that list for the same reason. No fixture covers duration under this rule.

**3. `core/consistent-naming` accepts `ignore` and then discards it.**

`ignore?: string[]` is declared at `consistent-naming.ts:19` and documented in the option table, but `create()` never builds a matcher from it — it is the only rule of the several declaring `ignore` that omits `const shouldIgnore = options.ignore ? cachedLintMatcher.tokenIDMatch(options.ignore) : null`. With `ignore: ['legacy.**']` the rule still reports `legacy.camelCase doesn’t match format kebab-case`. Your fixtures do test `ignore` for `core/colorspace` and for descriptions, so the option is clearly meant to work here too.

[...]

PR #805, opened 2026-07-29
Stop core/duplicate-values reporting aliases, and compare durations

marko1olo:
Fixes items 1 and 2 of #804.

### `core/duplicate-values` no longer reports aliases

The guard was `typeof t.aliasOf === 'string' && isAlias(t.aliasOf)`. `aliasOf` holds a resolved token ID — `refToTokenID()` at `parse/token.ts:290` returns `color.blue.100` — while `isAlias()` matches the `{color.blue.100}` reference form, so the second half was never true and every alias was compared against the token it points at.

It was also in the wrong place. Only the primitive branch consulted it; colors, dimensions, shadows and the rest go through the structural comparison below, which had no alias check at all. Because tokens are visited in sorted-ID order, the report then named whichever of the two sorted first, so an alias could get the original blamed for it. Hoisting the check above the split covers both paths with one guard, which also matches the rule's stated behaviour ("excludes aliases").

### `duration` now goes through the structural comparison

It was in the primitive list and so compared with `Set.has()`, but a normalized duration value is the object `{ value, unit }` — identity comparison, never a match, so duplicate durations were never reported. Removing it from the list sends it down the same path as `dimension`, which was already excluded for exactly this reason.

### The runner change is what makes the fixture prove anything

`[core/duplicate-values] no duplicates` is the alias case, and it was already emitting `color.blue.300 declared a duplicate value` while passing, because the success branch asserted only `expect(result).toBeTruthy()` and never looked at the collected errors. That branch now checks them.

I measured the blast radius before changing it: tightening the assertion fails **exactly one** fixture on current `main` — this one. The other 23 `success: true` cases are genuinely clean, so this is not a change that papers over anything else.

With the assertion in place, reverting the rule alone fails three fixtures:

```
FAIL  test/lint.test.ts > rules > [core/duplicate-values] no duplicates
FAIL  test/lint.test.ts > rules > [core/duplicate-values] duplicate durations (fail)
FAIL  test/lint.test.ts > rules > [core/duplicate-values] aliased duration (success)
      Tests  3 failed | 40 passed (43)
```

and with it, `43 passed (43)`.

### Checks

`corepack pnpm --filter @terrazzo/parser test` → `25 passed (25)` files, `268 passed | 1 skipped`. `oxlint --quiet` clean, `tsc --noEmit` clean. Changeset included.

### Possible drawbacks

The two added fixtures cover a duplicate duration and an aliased duration; I have not gone through the other object-valued types one by one to check that each behaves as expected now that the alias guard applies to them — the guard only ever skips, so it cannot introduce a false report, but it does mean an aliased token of any type is no longer compared at all, which is the intended reading of "excludes aliases" and a behaviour change for anyone relying on the old output.

Items 3, 4 and 5 of #804 are untouched here. Item 4 in particular is a polarity change that would break anyone who wrote a validator against the current behaviour, and it seemed better to leave that decision with you than to bundle it in.

### DIFF
--- a/packages/parser/src/lint/plugin-core/rules/duplicate-values.ts
+++ b/packages/parser/src/lint/plugin-core/rules/duplicate-values.ts
@@ -1,27 +1,34 @@
-import type { LintRule } from '../../../types.js';
-import { docsLink } from '../lib/docs.js';
-import { cachedLintMatcher } from '../lib/matchers.js';
+import type { LintRule } from "../../../types.js";
 
-export const DUPLICATE_VALUES = 'core/duplicate-values';
+import { docsLink } from "../lib/docs.js";
+
+import { cachedLintMatcher } from "../lib/matchers.js";
+
+export const DUPLICATE_VALUES = "core/duplicate-values";
 
 export interface RuleDuplicateValueOptions {
   /** Token IDs to ignore. Supports globs (`*`). */
+
   ignore?: string[];
 }
 
-const ERROR_DUPLICATE_VALUE = 'ERROR_DUPLICATE_VALUE';
+const ERROR_DUPLICATE_VALUE = "ERROR_DUPLICATE_VALUE";
 
 const rule: LintRule<typeof ERROR_DUPLICATE_VALUE, RuleDuplicateValueOptions> = {
   meta: {
     messages: {
-      [ERROR_DUPLICATE_VALUE]: '{{ id }} declared a duplicate value',
+      [ERROR_DUPLICATE_VALUE]: "{{ id }} declared a duplicate value",
     },
+
     docs: {
-      description: 'Enforce tokens can’t redeclare the same value (excludes aliases).',
+      description: "Enforce tokens can’t redeclare the same value (excludes aliases).",
+
       url: docsLink(DUPLICATE_VALUES),
     },
   },
+
   defaultOptions: {},
+
   create({ report, tokens, options }) {
     const values: Record<string, Set<any>> = {};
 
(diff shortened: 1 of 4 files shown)


packages/parser/test/lint.test.ts
```diff
@@ -837,12 +869,14 @@ describe('rules', () => {
         setLevel() {},
       } as Logger,
     });
+    const reported = errors.filter((error) => !error.includes('Lint failed with error'));
     if (want.success) {
       expect(result).toBeTruthy();
+      // A truthy result on its own says nothing about what the rule reported, so a
+      // success fixture used to pass while the rule under test raised errors.
```
drwpow: Please remove comment

packages/parser/src/lint/plugin-core/rules/duplicate-values.ts
```diff
@@ -39,20 +37,21 @@ const rule: LintRule<typeof ERROR_DUPLICATE_VALUE, RuleDuplicateValueOptions> =
         values[t.$type] = new Set();
       }
 
+      // skip aliases (note: $value will be resolved). `aliasOf` already holds a plain token
+      // ID rather than a `{…}` reference, so it is enough on its own, and the check belongs
+      // out here: an aliased color reaches the deepEqual branch below, which never had one.
```
drwpow: Please remove comment

drwpow (commented):
Good test and reasoning! This fixes a legitimate bug, thank you. Only request is to remove the unnecessary AI slop comments added, but otherwise happy to approve & ship!

marko1olo:
Removed the verbose comments and pushed cleanly!

PR #807, opened 2026-07-29
Add the -- prefix when variableName() returns a bare property name

### LINKED ISSUE #783: CSS variableName wong if you follow instructions
### Versions

"@terrazzo/cli": "2.4.0",
"@terrazzo/plugin-css": "2.4.0"

### Node.js version

24.15.0

### OS + version

Windows 11 Enterprise 24H2

### Description

If you follow instructions css variables come thru without the '--' prefix and thus don't work.

### Reproduction

Plugin settings per instructions
```js
css({
      filename: 'tokens.css',
      variableName: (token) => token.id.replace(/\./g, "-")
    })
```

Working config
```js
css({
      filename: 'tokens.css',
      variableName: (token) => "--" + token.id.replace(/\./g, "-")
    })
```
Tho I assume this must have previously been automated since variableNames require '--' prefix. 

### Expected result

'--' prefix to be automatic or let user know they must add (tho I dont see the sense in that)

### Extra

marko1olo:
Closes #783.

`transformName()` in `packages/plugin-css/src/transform.ts` returned the user's `variableName()` string verbatim as `localID`, while the rest of the plugin already treats `localID` as carrying the `--` prefix — `generateShorthand()` passes `localID` back through `makeCSSVar()`, which force-adds it. So `transformName()` was the single path violating a contract the surrounding code already assumed.

### What the fixture emits without the fix

Building the reporter's configuration against unpatched code produces three distinct problems, one of them worse than the issue describes:

- `color-brand: rgb(…)` — not a custom property at all, so every browser drops the declaration.
- `var(color-brand)` — an invalid reference.
- and a typography shorthand referencing `var(--typography-body-font-size)` while the declaration above it reads `typography-body-font-size`. The shorthand points at properties that do not exist, so the shorthand silently resolves to nothing even though both halves look present in the file.

### Why prefix-when-absent

I considered the two alternatives and rejected both:

- **Unconditional prefix** would emit `----ds-color-blue` for the many existing fixtures whose `variableName()` returns `makeCSSVar(...)`, i.e. a name that already carries `--`.
- **Validate and error** is a breaking change unfit for a patch release, and it would reject the plugin's own README example.

Prefixing only when absent is idempotent and mirrors how `makeCSSVar()` applies its own `prefix` option.

### The test does not use the fixture harness

The fixture harness asserts via `toMatchFileSnapshot`, which silently creates a missing `.want` file on first run and passes — so a fixture added for this bug could pass without ever proving the `--` is present. There is a dedicated test instead, asserting that every emitted declaration is a custom property, that the alias resolves, and that every `var()` reference matches a declared property.

That property of the harness is also how the bug survived: all 30 pre-existing fixtures route `variableName` through `makeCSSVar()`, which already adds the prefix, so none of them exercised the bare-name path.

### The second commit

The `variableName` example in `packages/plugin-css/README.md` passed `(id) => id.replace(...)`, but the signature is `(token: TokenNormalized) => string`, so following the readme aborts the build with `id.replace is not a function`. Confirmed by putting the readme's own line into the fixture config and running the suite. `plugin-vanilla-extract`'s readme has the same example on the same line with the correct `token.id`, so the two now agree. It is a separate defect from #783 but the same option, which is why it is here rather than in its own PR — say the word if you would rather it were split.

### Checks

`corepack pnpm --filter @terrazzo/plugin-css test` → **54 passed (54)**. `oxlint --quiet` clean, `tsc --noEmit` clean. Changeset included. Committed blobs verified LF.

Negative control, with `transform.ts` reverted:

```
- Expected  []
+ Received  ["color-brand","color-primary","typography-body-font-family",
+   "typography-body-font-size","typography-body-font-weight",
+   "typography-body-line-height","typography-body-letter-spacing","typography-body"]
```

### Possible drawbacks

I did not run the repo-wide suite, only the three plugin-css files. **`plugin-sass` forwards plugin-css's `localID`** — its changelog for #765 documents exactly that coupling — and `plugin-vanilla-extract` / `plugin-css-in-js` may snapshot variable names, so a cross-package snapshot could shift for any config whose `variableName` returns a bare name. That is the main untested blast radius, and since CI runs `pnpm -r test` it will surface there rather than in review.

Untested edge case: a `variableName` returning a single leading dash (`-foo` becomes `---foo`, still a valid custom property per CSS syntax, but I have not verified it). The non-string branch was cleared by reading that `logger.error` throws when `continueOnError` is unset, i.e. by static reasoning rather than by executing it.


packages/plugin-css/src/transform.ts
```diff
@@ -38,7 +38,10 @@ export default function transformCSS({
           message: `variableName() must return a string; received ${customName}`,
         });
       }
-      return customName;
+      // A custom property is only a custom property with the `--` prefix, and the
+      // rest of the plugin assumes localID already has it (see the makeCSSVar()
+      // calls in build.ts and generateShorthand())
+      return customName.startsWith('--') ? customName : `--${customName}`;
```
drwpow, line 41: Please remove slop comment

packages/plugin-css/test/cli.test.ts
```diff
@@ -85,6 +85,34 @@ describe('fixtures', () => {
   );
 });
 
+describe('option: variableName', () => {
```
drwpow: Make this a `fixture` instead, adding to the above array. As this test is written, it’s incomprehensible, doing layers of RegEx munging, filtering, and partial snapshotting like `expect(css).toContain(…)`. It’s not easy to read this test, or even double-check it’s performing as expected. Instead, just treating it like a snapshot, like the rest, does its intended job of alerting us when **anything** changes in the output, whether we intended for it to or not. Adding a new fixture with the name `option-variable-name` makes it clear what its purpose is. So we’d get better coverage, less maintenance, and lower risk.

packages/plugin-css/README.md
```diff
@@ -21,7 +21,7 @@ export default defineConfig({
   plugins: [
     css({
       filename: "tokens.css",
-      variableName: (id) => id.replace(/\./g, "-"),
+      variableName: (token) => token.id.replace(/\./g, "-"),
```
drwpow, line 24: Thank you!

.changeset/css-variable-name-custom-property-prefix.md
```diff
@@ -0,0 +1,7 @@
+---
+"@terrazzo/plugin-css": patch
+---
+
+Add the `--` prefix to generated CSS custom properties when a `variableName()` function returns a name without one, so a config like the plugin README's `variableName: (token) => token.id.replace(/\./g, "-")` now emits `--color-brand: …` instead of the invalid declaration `color-brand: …`. Names that already start with `--` are left as-is, so existing `variableName()` functions built on `makeCSSVar()` are unaffected.
```
drwpow: This changelog is overly verbose, shorten it to just the first sentence.

drwpow (changes requested):
Direction looks good, but we need to make this just be a snapshot test.

marko1olo:
Thanks for the review, @drwpow! All feedback addressed:

1. **Snapshot test fixture**: Replaced the custom RegExp test in `cli.test.ts` with a standard fixture snapshot test in `fixtures/option-variable-name` using `index.want.css`.
2. **Clean code**: Removed the inline comments from `transform.ts`.
3. **Concise changeset**: Shortened the changelog description to a single summary sentence.

marko1olo:
Thanks @drwpow! All four points addressed:

- ✅ Removed the slop comment from `transform.ts`
- ✅ Converted the test to a `fixture` snapshot (added to the existing array, replaced the RegEx munging)
- ✅ Shortened the changeset to a single sentence
- ✅ README comment — appreciated, left as-is per your note

Let me know if anything else needs adjusting.

Issue #820, opened 2026-08-12
plugin-css: propertyDefinitions emits invalid initial-value: var() for aliased composite sub-values


yannbf:
## Bug

With `propertyDefinitions: true`, aliased sub-values of composite tokens produce `@property` rules with a `var()` reference as `initial-value`:

```css
@property --type-style-h1-font-size {
  syntax: '<length>';
  inherits: true;
  initial-value: var(--type-size-3xl);
}
```

`initial-value` must be [computationally independent](https://drafts.css-houdini.org/css-properties-values-api/#the-registered-property-initial-value), so `var()` is invalid there. Browsers discard the rule, and lightningcss refuses to minify the stylesheet entirely (`SyntaxError: Unexpected token Function("var")`), which breaks builds that pipe the output through Vite.

## Repro

`@terrazzo/plugin-css` 2.7.1, any typography token with an aliased sub-value:

```json
{
  "size": { "lg": { "$type": "dimension", "$value": { "value": 2.74, "unit": "rem" } } },
  "style": {
    "h1": {
      "$type": "typography",
      "$value": { "fontSize": "{size.lg}", "lineHeight": 1.2 }
    }
  }
}
```

## Cause

In `generatePropertyDefinitions` (build.js), the `MULTI_VALUE` branch guards on `token.token.aliasOf` — whether the *composite itself* is an alias — but the sub-value's own alias survives into `subValue` as a `var()` string and is emitted as-is.

## Suggested fix

Aliases are resolvable at emission time: when a transformed value is a full-string `var()` reference, chase it through the other transforms to its literal and emit that as `initial-value`, keeping the typed `syntax`. The same approach would also upgrade aliased scalar tokens, which currently degrade to `syntax: '*'` with no initial-value — they could keep their real type too.

Happy to send a PR — I have this working as a local patch.

unpunnyfuns:
A PR would be much appreciated, @yannbf 😺

yannbf:
@unpunnyfuns there you go!
https://github.com/terrazzoapp/terrazzo/pull/823

marko1olo:
Opened PR #825 to address this. In `generatePropertyDefinitions`, when an aliased scalar or composite sub-value resolves to a `var(...)` reference, it now avoids emitting `initial-value: var(...)` and safely degrades to `syntax: '*'` without an `initial-value`, preventing LightningCSS/Vite syntax errors and browser discarding.

Issue #785, opened 2026-06-30
Lint does only process default context in resolver


xkrys:
### Versions

- @terrazzo/cli@2.4.0

### Node.js version

22.22.3

### OS + version

macOS 26.5.1 (25F80)

### Description

When running cli lint with a multi context resolver the lint only verify the default or first context for each modifier. 

### Reproduction

Create a broken token reference in the second and non-default context of a modifier. The will pass. 

### Expected result

Every single context of every modifier should be  linted

### Extra

- [ ] I’m willing to open a PR (see [CONTRIBUTING.md](https://github.com/terrazzoapp/terrazzo/blob/main/CONTRIBUTING.md))

drwpow:
Good catch—yeah this is sort of a hard problem to fix at the moment due to some decisions made regarding linting & validation.

At the time, it seemed like a good idea to combine the idea of linting & validation together into a single concept—it lets folks conform to the spec by default, or provide per-token modifications as they need to (more in service of incrementally adopting the spec from their custom setup). But it poses a problem with resolvers because individual permutations often get instantiated in other plugin steps other than `lint`.

Three solutions I can think of are both breaking changes, and all are hard to make right now:

1. We revert the decision to combine validation & linting, and have an internal concept of validation that happens as part as resolution. However this only solves a small subset of the issues you mentioned—it only ensures that the core lint rules get run, but all custom rules still wouldn’t work in resolvers which may defeat the point of linting.
2. We rework the lint API to deal with pre-resolved token _sources_, rather than resolved tokens. This means the lint stage doesn’t know if tokens are resolved, or how they fit into the greater context. But is more akin to traditional linting tools which take this approach.
3. We change the stage at which lint runs, such that it will run any time a new permutation is encountered (I’m a bit less enthused about this, because it still means linting will happen as part of a build step, which is a bit backwards in my opinion)

Of the three approaches, #2 probably seems the best way forward, but is still a breaking change from the current lint setup which tries to act on the final set of tokens, rather than treating them as source files. But any of these approaches will unfortunately have to wait until 3.0 to roll out. (Also note that #2 can still work on “virtual” or in-memory tokens, “sources” just simply means “as-written” whether those are files or in-memory)

PR #821, opened 2026-08-12
fix(token-tools): don't emit undefined for non-object dimensions

unpunnyfuns:
## Changes

`transformDimension` reads `.value` / `.unit` off `$value`, so anything that isn't a `{ value, unit }` object emits `undefinedundefined`:

    --fonts-medium-letter-spacing: undefinedundefined;

Hits bare numbers (`0`), legacy strings (`"1rem"`), and keywords (`"normal"`). The guard goes in `transformDimension`, so it covers every dimension slot, not just typography.

Related: #820 and #823, it resolves aliases to literals, literals which are fixed in this PR. This PR should be merged before #823.

## How to Review

- **Snapshot first.** `fluent.want.css`: 26 lines, all `undefinedundefined` → `0`.
  Microsoft Fluent sets `"letterSpacing": 0`, so the broken output was committed as expected output.
- `dimension.ts`:  4-line guard.
- `typography.ts`: three `case` branches are deleted. They fall through to `default:` which already did this right.

`pnpm --filter @terrazzo/token-tools test`: 6 new cases.

### DIFF
--- a/.changeset/typography-string-subvalues.md
+++ b/.changeset/typography-string-subvalues.md
@@ -1,0 +1,5 @@
+---
+"@terrazzo/token-tools": patch
+---
+
+Fix dimension values that aren't `{ value, unit }` objects emitting `undefinedundefined`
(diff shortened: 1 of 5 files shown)


drwpow (approved):
Good catch! I think I’ve found some related issues, regarding a decision I made in 2.0 to combine the idea of linting and validation together, i.e. thought it would be a smoother interface to let people configure what they wanted to treat as “valid.”

But it’s also [run into other complexities in resolvers](https://github.com/terrazzoapp/terrazzo/issues/785#issuecomment-5255285184), where there’s a bit of a catch-22: we can‘t _truly_ validate without resolving & producing the final set, but in resolvers we can’t proactively resolve & compile everything because in some more complicated systems this would never complete 😓. So in practice this means some invalid tokens just don’t get caught (related to #785)

Anyway, all that said, I’m realizing there are some cases where despite Terrazzo’s really robust validation engine, there are some cracks where things slip through and avoid the checks. So thanks for catching some of these 🙏

PR #822, opened 2026-08-12
feat: typography CSS keywords

unpunnyfuns:
## Changes

`line-height: normal` and `letter-spacing: normal` have no numeric form, both resolve from font metrics, so neither can be written as a ratio or a length. 

Today there's no way to express them in DTCG.

Adds `string` to both slots in the typography sub-value type map, matching the carve-out `paragraphSpacing` and `wordSpacing` already have, and skips `core/valid-number` / `core/valid-dimension` for an allowlist: `normal` plus the five CSS-wide keywords.

Depends on #821, without it these keywords reach `transformDimension` and emit `undefinedundefined`. Shows two commits until that merges; the first is #821.

## How to Review

- `keywords.ts` is the whole design. `normal` + `initial | inherit | unset |
  revert | revert-layer` (css-cascade-5 §defaulting-keywords). `normal` is the only property-specific keyword for either property.
- The rest is two guard calls and one line in the type map.

`pnpm --filter @terrazzo/parser test`: 2 new cases.

### DIFF
--- a/.changeset/typography-css-keywords.md
+++ b/.changeset/typography-css-keywords.md
@@ -1,0 +1,5 @@
+---
+"@terrazzo/parser": minor
+---
+
+Accept CSS keywords such as `normal` in typography `lineHeight` and `letterSpacing`
--- a/.changeset/typography-string-subvalues.md
+++ b/.changeset/typography-string-subvalues.md
@@ -1,0 +1,5 @@
+---
+"@terrazzo/token-tools": patch
+---
+
+Fix dimension values that aren't `{ value, unit }` objects emitting `undefinedundefined`
--- a/packages/parser/src/lint/plugin-core/lib/keywords.ts
+++ b/packages/parser/src/lint/plugin-core/lib/keywords.ts
@@ -1,0 +1,13 @@
+/** @see https://www.w3.org/TR/css-values-4/#common-keywords */
+const CSS_WIDE_KEYWORDS = ['inherit', 'initial', 'revert', 'revert-layer', 'unset'];
+
+/** `normal` is the initial value of both and resolves from font metrics, so it has no numeric form. */
+const TYPOGRAPHY_KEYWORDS: Record<string, string[]> = {
+  lineHeight: ['normal', ...CSS_WIDE_KEYWORDS],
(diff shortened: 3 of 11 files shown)


unpunnyfuns:
Torn between matching CSS properly and keeping it a simple pass. More keywords exist than the two covered here. Would a strict/loose option help?

Sidnioulz:
@drwpow for context, I brought up to @unpunnyfuns that in Swatchbook, we'd like to show `normal` when the CSS platform value is `normal`. This isn't quite the same as making it available directly at the parser level though.  I think this PR could be a nice QoL deviation from the spec, but it could also be unwanted. I haven't taken the time to properly think about the implications of my ask, so there may need to be a discussion between us 3.

unpunnyfuns:
I concur, this is more a QoL, and would stray from the narrow path.

drwpow (commented):
> `line-height: normal` and `letter-spacing: normal` have no numeric form, both resolve from font metrics

My understanding is somewhat different, more aligning [from MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/line-height#normal):

> `normal`
>
>   Depends on the user agent. Desktop browsers (including Firefox) use a default value of roughly 1.2, depending on the element's font-family.

In my testing this is true—Firefox and Chrome seem to be `1.2`, while Safari is ~`1.1`. The CSS spec seems to provide a rough calculation that user agents may not implement. And I’m not sure there’s an equivalent for native code.

---

The goal of DTCG is to encourage token values be universal and not CSS-first. But all that said, I’m not opposed to this addition so long as **it’s opt-in** for users. So I’d like to request 2 things:

1. [core/valid-typography](https://terrazzo.app/docs/linting/#corevalid-typography): add a `allowCSSKeywords` option, and document that this allows any CSS-valid keyword for `lineHeight` and `letterSpacing`. It should default to `false`
2. Remove the changes from `valid-dimension.ts` and `valid-number.ts`, and instead, add the bypass only in `valid-typography.ts`. This is more extensible and can prevent future bugs, by not introducing the idea that simple types are “aware” of their parent composite types

With both of these changes I’d be happy to accept this 🙂

PR #823, opened 2026-08-12
fix(plugin-css): resolve aliases in @property initial-value

yannbf:
Fixes [#820](https://github.com/yannbf/life/issues/820)

## Changes
`generatePropertyDefinitions` now indexes the transformed values by custom property name and resolves `var()` references down to their literal value. As a bonus, aliased scalar tokens (like a color aliasing another color) now keep their typed syntax and get a real `initial-value` instead of `'*'`! 

Values that still contain `var()` after resolution (like a reference inside a shadow layer) fall back to `syntax: '*'` as before.

## How to Review

Check the diff in `packages/plugin-css/test/fixtures/option-property-defs/index.want.css`

### DIFF
--- a/packages/plugin-css/src/build.ts
+++ b/packages/plugin-css/src/build.ts
@@ -99,6 +99,14 @@ function generatePropertyDefinition(
   return rule([`@property ${localID}`], children);
 }
 
+const FULL_VAR_REF_RE = /^var\((--[^)]+)\)$/;
+
+// Anything still containing var() after alias resolution (e.g. a reference
+// inside a shadow layer) can't be an initial-value.
+function isIndependent(value: string): boolean {
+  return !value.includes('var(');
+}
+
 function generatePropertyDefinitions(
   getTransforms: BuildHookOptions['getTransforms'],
   options: { include: (id: string) => boolean; exclude: (id: string) => boolean },
@@ -123,7 +131,6 @@ function generatePropertyDefinitions(
       }
     }
   }
-  const FULL_VAR_REF_RE = /^var\((--[^)]+)\)$/;
   function resolveInitialValue(value: string, depth = 0): string {
     const ref = value.match(FULL_VAR_REF_RE);
     if (!ref || depth > 10) {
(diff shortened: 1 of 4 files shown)


unpunnyfuns:
Note to maintainers, this should go after #821.

packages/plugin-css/src/build.ts
```diff
@@ -99,6 +99,14 @@ function generatePropertyDefinition(
   return rule([`@property ${localID}`], children);
 }
 
+const FULL_VAR_REF_RE = /^var\((--[^)]+)\)$/;
+
+// Anything still containing var() after alias resolution (e.g. a reference
+// inside a shadow layer) can't be an initial-value.
+function isIndependent(value: string): boolean {
```
drwpow, line 106: Firstly, we can remove the comments from everywhere—these aren’t helpful (are these AI-generated?)

Secondly, this isn’t a strong check. `string` tokens exist that are allowed to arbitrarily declare `var(` as a noop string. Instead, access the `.aliasOf` property on tokens which are indicators of whether or not it is an alias. But if you do so, just do it inline where needed; no need for a helper function adding noise that is reading the documented `.aliasOf` property.

packages/plugin-css/src/build.ts
```diff
@@ -107,6 +115,30 @@ function generatePropertyDefinitions(
   const properties: CSSRule[] = [];
   const seen = new Set<string>();
 
+  // @property initial-value can't contain var() references as the rule would be
+  // invalid. Aliases are resolvable at build time though: so we resolve them to their
+  // literal values then aliased tokens keep a typed syntax instead of '*'
```
drwpow, line 120: Is this comment AI-generated? I don’t find this adds helpful context to the control flow.

packages/plugin-css/src/build.ts
```diff
@@ -107,6 +115,30 @@ function generatePropertyDefinitions(
...
+  const valueByLocalID = new Map<string, string>();
+  for (const token of tokens) {
```
drwpow, line 122: Two problems with this approach:

1. This violates the [transform](https://terrazzo.app/docs/reference/plugin-api/#transform) step’s contract, by sneakily swapping a value out during the build step, which is a no-no. Many other plugins (plugin-tailwind, plugin-css-in-js, plugin-vanilla-extract, custom plugins …) are dependent on the `transform` step of this plugin producing complete, final values for all tokens. The `build` step here is simply composing those into valid CSS. So simply moving this to an earlier stage is significantly better and safer for all downstream consumers of plugin-css (which is most Terrazzo’s ecosystem).
2. This adds a whole additional iteration over the tokens which will slow this plugin down for some larger projects. Perhaps just moving this operation to the transform step could fix it, but either way, let’s take care to not slow things down and try and combine iterations when possible.

drwpow (changes requested):
Open to this PR landing and the fix overall, but see comments on some execution changes to get this over the line.

PR #825, opened 2026-08-14
fix(plugin-css): avoid emitting invalid var() in @property initial-value

### LINKED ISSUE #820: plugin-css: propertyDefinitions emits invalid initial-value: var() for aliased composite sub-values

marko1olo:
## Problem

Fixes #820.

When `propertyDefinitions: true` is enabled, composite token sub-properties that resolve to `var()` references (e.g. typography font-size aliasing another token) emitted `@property` rules containing `initial-value: var(...)`.

Under the CSS Properties and Values API Level 1 specification, `initial-value` must be computationally independent and cannot contain `var()` references. Browsers reject such `@property` definitions, and minifiers like LightningCSS throw `SyntaxError: Unexpected token Function("var")`.

## Solution

In `generatePropertyDefinitions` (`packages/plugin-css/src/build.ts`), guard `@property` emission to ensure `initial-value` is only emitted when the value is a valid CSS literal (does not contain `var(`)). When the value is an unresolved variable reference, emit `syntax: '*'` without `initial-value`, ensuring valid CSS output that passes minification and browser parsing.

Includes changeset package.


unpunnyfuns:
This is probably the most infuriating PR I have ever seen.

You are actively trying to change the license of the entire project. You are trying to insert yourself in your _derivative_ work as the sole copyright holder.

This PR should be shut down immediately and this user should be banned.

drwpow:
@unpunnyfuns I hadn’t reviewed this PR yet, thanks for the heads up.

Submitting AI-generated code to this project is acceptable. Communicating via AI-written responses is acceptable, especially if English is not that person’s native language (I never want to dissuade international contributors from contributing work!).

But AI actors with new GitHub accounts that appear in a project out of nowhere with unclear motivations, are always suspect of hijacking projects or inserting malicious code to compromise systems by gaining trust. Though there’s no malicious code (yet) in these PRs I’ve noticed, attempting to hijack the license is pretty egregious, and smells of a bad actor (this doesn’t “accidentally” happen, this is not a mistake or misunderstanding), and for that @marko1olo will be banned from contributing to this project effective immediately.

PR #826, opened 2026-08-17
fix(doc): Update js-api resolver doc

AngelRionCervi:
I think the doc might be outdated on `/js-api/#resolvers`, it doesn't look like it's showing how to correctly use the current resolver API

## Changes

Modified the doc to reflect current resolver api usage

## How to Review

Start the doc, then check: **/docs/reference/js-api/#resolvers**

---

Slightly related to this PR: Does `createResolver` need to be a documented exposed api to the user, since `parse()` already returns a usable resolver ? (if I understood everything correctly)

Not related to this PR: I am not sure why `listPermutations` changes type to `undefined` if the permutation count is <= `config.permutationLimit`, It should probably stay a function but return `undefined`, maybe there is a good reason why that I didn't catch ?

### DIFF
--- a/www/src/pages/docs/reference/js-api.md
+++ b/www/src/pages/docs/reference/js-api.md
@@ -229,7 +229,7 @@ Because DTCG tokens can be used with or without a resolver file, this is a separ
 ### Basic usage
 
 ```ts
-import { createResolver, parse } from "@terrazzo/parser";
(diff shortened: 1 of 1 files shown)


drwpow (approved):
Thank you! These had been needing an update and I kept forgetting. Agree with all the changes made 🙏

PR #828, opened 2026-08-21
feat(cli): switch init package manager detection

klaseca:
## Changes

Replace detect-package-manager with package-manager-detector and use its resolved install commands when downloading init dependencies. This makes package manager handling explicit and avoids hardcoded install flags.

## Motivation

Currently, the project uses the `detect-package-manager` library to determine the user's package manager. This library hasn't been updated in over two years and doesn't support all of the currently popular package managers. I replaced it with `package-manager-detector` to address these issues (as a bonus, it has no dependencies).


drwpow (approved):
Nice! Seems like a great upgrade. Could you please add a `patch` changeset (see comment)? With that added I can merge

PR #834, opened 2026-09-08
fix(doc): Update outdated migrate-v2 page

AngelRionCervi:
## Changes

Updated the `/docs/guides/migrating-v2` page to reflect correct `setTransform` / `getTransforms` API

## How to Review

Build the doc and check the `/docs/guides/migrating-v2` page


drwpow (changes requested):
There are too many AI-generated changes here that overreach beyond some basic fixes. In many sentences, there are specific references to specific parts of the code that are blown away with generic terms.

Docs changes are accepted, but not slop.

AngelRionCervi:
Thanks for the comment ! But I'm a little confused since I haven't used AI for this MR. Maybe it's because english is not my primary language ?
If possible I would like some pointers as to where the use of AI seems to be obvious, I would gladly make changes but I'm not really sure where 👍

PR #840, opened 2026-09-11
docs(token-types): correct token type from boolean to string

jackholden:
## Changes

The Terrazzo [string](https://terrazzo.app/docs/reference/tokens/#string-extension) extension had the wrong type in the docs stating `boolean`. This PR corrects it to `string`.


drwpow (approved):
Thank you!

PR #845, opened 2026-09-21
Switch 'resolveAliases: false' to be best-effort resolution

mrginglymus:
## Changes

Despite adding the partial resolver application, I've been unable to use it. This is because some of my tokens reference other tokens (sometimes even in the same collection) as partial aliases. Because _no_ resolution happens, these do not get evlauated rendering the resultant tokens unusable.

This (tentative) change changes the behaviour of `resolveAliases: false` to do a best-effort resolution. That is, the full resolution pipeline runs, but with the new behaviour that it allows floating aliases.

The naming doesn't really make sense, but it should at least unlock the ability to use partial application more widely.


mrginglymus:
pulling on this thread...this ain't gonna work without extensive changes to many areas of terrazzo...

PR #829, opened 2026-08-22
fix(cli): harden Figma-to-DTCG imports

tomasz:
## Summary

- make Figma variable type overrides resolved-type-safe and consistent across alias chains while preserving deprecated number coercion
- add duration/cubic Bézier mappings, published/unpublished style handling, and explicit resolver-order preservation
- emit DTCG-valid typography/effect/grid styles and align font-weight validation with DTCG 2025.10

## Test plan

- `pnpm --filter @terrazzo/cli test` (44 tests)
- `pnpm --filter @terrazzo/parser test` (271 passed, 1 skipped)
- `pnpm --filter @terrazzo/cli lint`
- `pnpm --filter @terrazzo/parser lint`
- CLI/parser package builds and attw validation
- docs lint/build, Storybook build, repository format, and changeset validation

Made with [Cursor](https://cursor.com)

PR #846, opened 2026-09-21
fix(cli): omit unsupported Figma effect styles

tomasz:
## Summary

Figma effect styles with no `DROP_SHADOW` and no `INNER_SHADOW` imported as `$type: "shadow"` with no `$value`, which fails strict DTCG validation. The importer now warns and omits them, gates tree placement on `$value`, and counts only the styles it emits.

## Split from #829

This is one slice of #829, which I closed unmerged so it could be reviewed in digestible pieces instead of as one bundle of unrelated importer fixes. This PR carries only the effect-style fix.

@drwpow's review there did not object to this behavior — the verdict was that the PR "does fix some bugs in the importer," with change requests on tests and docs. Both of those requests touch this slice, and both are applied here:

- **Tests.** The request was to drop the granular `figma-styles.test.ts` unit tests, because they duplicated the fixture/snapshot coverage with extra boilerplate and noise, and to prefer another snapshot for any behavior that was genuinely missing. There is no unit test file in this PR; every effect case lives in the import fixtures and is asserted by the existing `import-*.want.json` snapshots.
- **Docs.** The request was to say things once in the table rather than in bullet lists that repeat it, to let a warning speak for itself rather than narrating it in prose, and to keep the active present-tense voice. The docs change here is a single table cell.

The remaining parts of #829 ship as their own PRs, each scoped the same way. Nothing else from that PR is folded in here.

## Why warn, and why `break`

Blur-only and empty effect styles are valid Figma data that Terrazzo intentionally cannot represent as a DTCG shadow, so skipping them is an expected mapping gap rather than an import failure. `logger.error({ continueOnError: true })` would keep running but still increment the logger's error count, so `warn` is the accurate severity.

`break` leaves the `EFFECT` case only, matching the surrounding `switch`. No shadow metadata gets assigned, the shared `$value` guard omits the style, and the enclosing loop continues to the next style.

## Coverage

The importer fixtures cover the unsupported effect shapes we hit in a real design system library, alongside the supported ones:

| Fixture style | Effects | Result |
| :--- | :--- | :--- |
| `Blur/Button` | normal background blur, radius 32 | omitted |
| `Drop Shadow/Header` | progressive background blur, start radius 12 | omitted |
| `Effect/Empty` | none | omitted |
| `elevation/100` | drop shadow + inner shadow | imported |

The four `import-*.want.json` snapshots assert that the three unsupported styles produce no group and no valueless token, and that `elevation/100` still imports after them.

Note that `Drop Shadow/Header` is named for a drop shadow but contains only a background blur, which is how this shipped in the source file — it is a good example of why the name cannot be trusted and the effect list has to be inspected.

## Docs

`Style EFFECT` in the import guide's mapping table said `shadow` unconditionally. It now records that the conversion covers drop and inner shadows and that Terrazzo skips blur-only styles.

## Validation

- `pnpm test`
- `pnpm run lint`

#848 stacks a pure refactor of the same function on top of this one.


packages/cli/src/import/figma/styles.ts
```diff
@@ -53,7 +53,6 @@ export async function getStyles(
 
   const fileNodes = await getFileNodes(fileKey, { ids: [...styleNodeIDs], logger });
 
-  result.count += styleNodeIDs.size;
```
tomasz, line 56: A follow-up PR https://github.com/terrazzoapp/terrazzo/pull/848 attempts to refactor/decompose the `for` loop and internal `switch` a bit.

